PR #20, opened 2018-07-29
Issue/18/embed illustration

aimalz:
I added a readme for the code as distinct from the project, embedded the code structure doc in it, removed the old readme that was hanging around in the top level of the repo, and removed the static image I'd put in the docs.


cavestruz:
Looks good - can see illustration in clmm readme.  Thanks!

Issue #28, opened 2018-07-29
Need a license file in the repo


matthewkirby:
The license file is super important in the repo. The way I have had it explained to me is that without it, we can "technically" get other people in legal trouble for using our code. We can avoid this just by adding one of the licenses to the repo. Is there one that desc requires/recommends?

cavestruz:
Check on restrictions on license based off of clmassmod (legally, what can people do with the code?)  Check if DESC has requirements.

drphilmarshall:
The pub board has provided some guidelines at
https://confluence.slac.stanford.edu/display/LSSTDESC/Publication+Board

>

PR #34, opened 2018-07-30
Added a license

matthewkirby:
Added the BSD 3-clause license and added a line to the readme to designate this as a desc project.

This is the license that is default suggested by the desc publication policy.

If our code depends on any software that uses the GPL license, we will need to swap to that. But the current state of the repository does not.


matthewkirby:
This is referencing issue #28

aimalz:
closes #28

PR #36, opened 2018-07-30
Issue/1/easyinstall

aimalz:
I'm attempting to set up installation via easyinstall and can more easily see if Travis-CI is failing if there's an open PR.  Hopefully this will be resolved quickly!


aimalz:
The `setup.py` works now, but the tests of the `models` package are failing.

Issue #39, opened 2018-07-30
Need to simplify the parameter object in workshop goal


matthewkirby:
For the workshop minimal example, we will not have the Parameter object implemented. We need to update existing tests in master to expect floats.

matthewkirby:
This issue is for a simplified implementation of model parameters. Parameters are being treated as dictionaries for now and this issue is resolved.

The issue and decision of how we treat parameters in the long run is still open in issue #22

PR #42, opened 2018-07-31
Changed parameter to be a dictionary for now. This should be reevalua…

matthewkirby:
…ted after the hackshop

Removed some commented out crap from the `setup.py` merge and stopped using the Parameter object in Model.


matthewkirby:
Upon merging this PR, issue #39 can be closed.

Issue #43, opened 2018-07-31
New module structure obliterated docs!


matthewkirby:
Docs straight up crash on imports.

Need to get apidoc running automatically so that this can be tested for!

Issue #1, opened 2018-06-01
Preliminary infrastructure


aimalz:
We need documentation and continuous integration ahead of the workshop.  This requires a hello world jupyter notebook, a .travis.yml and setup.py, and autodoc/Sphinx set up.

aimalz:
@matthewkirby is setting up Travis.  @aimalz is setting up Sphinx.

matthewkirby:
I want to open a PR for branch . The problem is that I merged it with issue/4/galaxycluster_outline so that there were docstrings to test against. Is the fix for this just to delete the galaxycluster.py I merged with and PR into master?

matthewkirby:
Whoops, the branch is testadddocstring. Sorry for the name =X it preceeded the explanation of branch names

msimet:
No, don't delete the file. The problem is that git would know that the
delete-file commit came after the commits that added the file, so it could
do funny things when you try to merge the other branch later. I would
suggest one of two options:

1. Just make the PR as is, but don't have anyone review or merge it until
the branch you merged in has gone through PR.
2. Use git log or https://github.com/LSSTDESC/CLMM/commits/testadddocstring
to find and copy the SHAs for all of the commits you made to this branch.
Then, make a new branch from master and 'git cherry-pick' all of the
commits onto the new branch in order, and do a PR from there.

There are probably other solutions too, but those are the two easiest I can
come up with off the top of my head!

matthewkirby:
Thanks for the suggestions. I issued a PR for the read the docs files.

cavestruz:
- Need automatically build the readthedocs
- Need to fix class documentation Issue #15

matthewkirby:
Things that still need to be done before this can be closed

- [ ] Readthedocs
    - [ ] Autoapi needs to be run automatically rather than manually.
- [ ] Travis-ci
- [ ] License (Issue #28)

PR #45, opened 2018-07-31
Issue/1/read-the-docs/0.0.2 and a lot of other things

matthewkirby:
This fix addresses issues #43 and the last remaining goal of issue #1 and after merging we will be able to close both issues.

__Changes__
- sphinx-autoapi is now run automatically by RTD to completely compile docs remotely
- I removed the directories and files that were used to do this locally from the repo
- I changed the authorship (as printed on the docs) of the repo back to LSST DESC Clusters Working Group.
- The docs no longer crash as they have been doing in master
- Added a docs status badge to the README
- Swapped from numpydoc to napoleon. Napoleon works natively on RTD whereas numpydoc requires some tricks so it wasn't worth it
- Fixed some documentation formatting problems in galaxycluster.py
- Did some "python-fu" (?) to get a docstring on GCData
- Commented out most of the things that travis installs. It was installing a lot of things that it was not using.
- Travis now fails on errors from sphinx
- Travis now fails on warnings from sphinx. This was done since most warnings are caused by malformed docstrings which do not render in docs correctly.
- Removed `clmm/inferrer/__init__.py` and `clmm/metrics/__init__.py`. These files were causing problems since they were compiling these directories into a subpackages that had exactly zero code in them causing sphinx warnings. This, however, also deleted the `inferrer` and `metrics` directories since they had no files in them.
- Set up pylint in travis but commented it out. It needs a little more configuration but our code in master has a ton of things that need to be changed, some of it is a little nontrivial/might not be an issue. It still requires more thought. Regardless, this should be pushed to an enhancement for after our minimal example is running.
- Added some module docstrings

__Questions__
- Should we put dummy files into `inferrer` and `metrics` so that we can have the empty directories exist or is it okay that they have vanished for now?
- I also noticed that our test are now in `root/tests/` but there is a single test in `root/clmm/tests/`. I tried moving it but I couldn't get the tests to pass once I moved it so I left it behind.


docs/conf.py
```diff
@@ -170,4 +172,18 @@
 ]
 
 
+# -- Options for Napoleon-------------------------------------------------
+napoleon_use_ivar = True
 
+# -- Options for Autodoc--------------------------------------------------
+def run_apidoc(_):
+    from sphinx.apidoc import main
```
cavestruz: It would be nice to have documentation showing how/why this gets autodoc working  (e.g. why certain flags are chosen, etc.) in case things need to be changed in the future.

clmm/models/__init__.py
```diff
@@ -1,2 +1,3 @@
+"""Modules that build off of the Model superclass"""
 from .model import *
```
cavestruz: Let's call this a temporary fix.  But, we will want to avoid this in the future to prevent potential namespace conflicts and also to make it easier to read through the code and trace through.

PR #54, opened 2018-07-31
Fix imports of galaxycluster

m-aguena:
Fix import in galaxycluster, datatypes
Put test_galaxycluster,py and test_shear_azimuthal_averager.py in tests
Fix bug in add and remove in galaxycluster.


m-aguena:
made it backwards

Issue #56, opened 2018-07-31
core directory creation


cavestruz:
Same level as models, summarizer.  Contains datatypes and weak_lensing_functions

PR #57, opened 2018-07-31
added core directory structure.  Things need to be added to datatypes

cavestruz:
#56


matthewkirby:
Travis tests are failing because you are not allowed to have empty directories. This directory structure should be

`root/clmm/core/`
`root/clmm/core/__init__.py`
`root/clmm/core/weak_lensing_functions.py`
`root/clmm/core/datatypes.py`

PR #59, opened 2018-07-31
improve documentation of GCData

m-aguena:
Specified what `creator` and `specs` should look like.


clmm/datatypes.py
```diff
@@ -12,14 +12,17 @@
...
     """
-    GCData: A namedtuple tying values with units to the metadata of where the values came from and thus how the values are used
+    GCData: A namedtuple tying values with units to the metadata of where the values came from
+    and thus how the values are used
 
     Parameters
     ----------
     creator: string
-        Type of object (i.e. model, summarizer, inferrer) that made this data
+        Type of object that made this data
```
matthewkirby: This should be the super class of the object that made the data.

m-aguena: done!

PR #61, opened 2018-07-31
Issue/58/ultra super class

matthewkirby:
Developed a super mega ultra class for every class outside of core to inherit. Contains an attribute that the manager class will read.


tests/test_clmmbase.py
```diff
@@ -0,0 +1,70 @@
...
+    class_list = list(set(class_list) - set(skip_classes) - set(extern_classes))
+
+    # Print the list of classes being analyzed. You will only see output
+    # on assertion failures
+    print(class_list)
+
+    # Assert that we have not added a new class to the api. If this fails, we
+    # need to instantiate an example of these objects below and test that
+    # ask_type has been set to a reasonable value
+    assert len(class_list) == 2
```
cavestruz: This should not have to be modified with each new class.

tests/test_clmmbase.py
```diff
@@ -0,0 +1,70 @@
...
+
+    t1 = clmm.CLMMBase()
+    t1.ask_type = ['something']
+    np.testing.assert_array_equal(t1.ask_type, ['something'])
+
+    t2 = clmm.CLMMBase()
+    np.testing.assert_raises(TypeError, t2.ask_type, 3.14159)
+
+
+def test_new_classes():
```
cavestruz: I like this check, but should be more automatic, and lets people know if they have to inherit this.

matthewkirby: done

matthewkirby:
I rewrote the tests to be automatic. Let me know if there are more changes that are requested.

PR #66, opened 2018-10-05
Issue/15/fake data

combet:
MockData class to generate mock background galaxy catalogs including:
* the ideal case: no redshift error, no shape noise
* the non-ideal case: optionally include redshift error and/or shape noise. If a redshift error is requested, the catalog also includes the associated Gaussian pdz.

At the moment, all background galaxies have the same redshift (with some error if requested).

This comes with a demo python notebook in the "example" directory.

Issue/15/fake_data could be closed for the time being.

### DIFF
--- a/clmm/__init__.py
+++ b/clmm/__init__.py
@@ -4,3 +4,4 @@ from __future__ import absolute_import
 from .models import *
 from .summarizer import *
 from .core import *
+from .mock_data import *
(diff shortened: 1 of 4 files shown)


cavestruz:
Made some minor changes to the ipynb demo, also created a new issue related to this so we can augment the mock data to take in src redshift distributions.

PR #69, opened 2018-10-22
Issue/23/manager

m-aguena:
first working version of the manager.


clmm/core/manager.py
```diff
@@ -0,0 +1,112 @@
...
+class Manager():
+
+    '''
+    Makes the interactions between GalaxyCluster and the other clmm objects
+    '''
+
+    def __init__(self):
+        '''
+        '''
+        pass
```
matthewkirby: Should this `pass` be here? There is code after it

clmm/data__types.py
```diff
@@ -0,0 +1,102 @@
+"""
```
matthewkirby: This file name has two underscores? Is this intended?

matthewkirby: It also has the same name as clmm/core/datatypes.py

matthewkirby: Actually, was the file just moved? What happened to the one in clmm/core?

m-aguena: @matthewkirby as far as I could find out, the file clmm/core/datatypes.py is the one being imported by all other modules. I dont know why clmm/data__types.py exists.
@combet do you know what is this purpose of this file? Is it actually being used? I could not find an import of it in any other part.

clmm/core/manager.py
m-aguena: removed

clmm/data__types.py
cavestruz: @matthewkirby  -  I think all requested changes were made.  It looks like you have to approve the requested changes(?)

PR #94, opened 2019-03-02
Issue/92/interaction manager galaxy cluster gc data

m-aguena:
I managed to make manager.appy run a dummy function. Had to create a distinction between func_spec and input_specs, and fixed a bug in manager._unpack.


matthewkirby:
I can take a look on Wednesday. Sounds exciting though!

aimalz:
The manager is now ready for a non-dummy test with CCL's NFW profile as a model!

PR #98, opened 2019-03-02
Issue/93/add print to galaxy cluster

m-aguena:
Added new nice print to GalaxyCluster object and cleaned print of 'no data found with these lookup_specs' for necessary cases only.

### DIFF
--- a/clmm/core/datatypes.py
+++ b/clmm/core/datatypes.py
@@ -86,20 +86,15 @@ def find_in_datalist(lookup_specs, datalist, exact=False):
     -------
     found: list
         List of clmm.GCData object data with required creator and set of specs
-    None
-        If no objects are found
     '''
     found = []
     for data in datalist:
         if check_subdict(lookup_specs, data.specs) :
             found.append(data)
-    if len(found) == 0:
-        print('no data found with these lookup_specs')
     else:
         if exact:
             for match in found:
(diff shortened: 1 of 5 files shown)


clmm/core/datatypes.py
```diff
@@ -86,20 +86,15 @@ def find_in_datalist(lookup_specs, datalist, exact=False):
...
-    None
-        If no objects are found
     '''
     found = []
     for data in datalist:
         if check_subdict(lookup_specs, data.specs) :
             found.append(data)
-    if len(found) == 0:
-        print('no data found with these lookup_specs')
     else:
```
maho3: Is this else supposed to have a matching if?

tests/test_manager.py
```diff
@@ -38,10 +38,10 @@ def test_pack() :
     tst.assert_equal(manager_guy._pack(func_test, test_spec, []), test_data_out)
 
 def test_unpack() :
-    tst.assert_equal(manager_guy._unpack(test_gc, func_test, test_spec), test_data)
+    tst.assert_equal(manager_guy._unpack(test_gc, func_test, test_spec), test_data.values)
 
 def test_apply() :
-    manager_guy.apply(test_gc, func_test, test_spec)
+    manager_guy.apply(test_gc, func_test, test_spec, test_spec)
```
maho3: Copy of the same argument?

Issue #104, opened 2019-07-12
Fix Sphinx after #94


aimalz:
It looks like PR #94 broke Sphinx even though it passed Travis CI tests before merging into master.  @matthewkirby @msimet Would you (or anyone else of course) mind looking into what happened?

matthewkirby:
Hmm it actually looks like travis is failing on master due to a sphinx error. I will take a look

matthewkirby:
The error I see on Travis is
```
0.75s$ make -C docs/ html
make: Entering directory `/home/travis/build/LSSTDESC/CLMM/docs'
Running Sphinx v2.1.2
making output directory... done
Exception occurred:
  File "/home/travis/build/LSSTDESC/CLMM/docs/conf.py", line 181, in run_apidoc
    from sphinx.apidoc import main
ModuleNotFoundError: No module named 'sphinx.apidoc'
The full traceback has been saved in /tmp/sphinx-err-0bbpm2pj.log, if you want to report the issue to the developers.
```
I pulled the new changes to master and ran `make -C docs/ html`.
```
09:35 AM  matthewkirby /Users/matthewkirby/research/CLMM/CLMM master
$ make -C docs/ html
Running Sphinx v1.6.6
loading pickled environment... done
Creating file /Users/matthewkirby/research/CLMM/CLMM/docs/api/clmm.galaxycluster.rst.
Creating file /Users/matthewkirby/research/CLMM/CLMM/docs/api/clmm.mock_data.rst.
Creating file /Users/matthewkirby/research/CLMM/CLMM/docs/api/clmm.rst.
Creating file /Users/matthewkirby/research/CLMM/CLMM/docs/api/clmm.core.clmmbase.rst.
Creating file /Users/matthewkirby/research/CLMM/CLMM/docs/api/clmm.core.datatypes.rst.
Creating file /Users/matthewkirby/research/CLMM/CLMM/docs/api/clmm.core.manager.rst.
Creating file /Users/matthewkirby/research/CLMM/CLMM/docs/api/clmm.core.weak_lensing_functions.rst.
Creating file /Users/matthewkirby/research/CLMM/CLMM/docs/api/clmm.core.rst.
Creating file /Users/matthewkirby/research/CLMM/CLMM/docs/api/clmm.inferrer.collections.rst.
Creating file /Users/matthewkirby/research/CLMM/CLMM/docs/api/clmm.inferrer.importance_sampling.rst.
[...]

PR #105, opened 2019-07-16
Fixing issue #104, sphinx broken in master

matthewkirby:
In late April, Sphinx deprecated versions 1.X and moved to 2.X. As we are not specifying a version of sphinx to use, Travis used the most recent version (v2.1.2) which then crashed and caused Travis to fail.

While Travis was failing, the docs looked good on readthedocs (rtd). This was because rtd was set to use the CPython 2.X interpreter. Since sphinx v2.X no longer has Python 2.X support, rtd defaulted to sphinx 1.X and compiled fine.

I have updated our sphinx files to be compatible with sphinx 2.X. I have also added the version number for sphinx in the travis config file to avoid this problem in the future.

Note: Sphinx 2.X generates a warning for the rtd template. A feature in this template will be deprecated in the future (shouldn't be an issue since we now specify v2.1.2 for sphinx). The developers of this template have fixed this and it will be included in the future v0.44. The most recent public release is v0.43 so we will have to live with this warning for a bit.

See the docs here: https://clmm.readthedocs.io/en/issue-104-sphinx-broken-in-master/api.html

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-docs/_build/*
 build
 dist
 *~
@@ -6,5 +5,9 @@ __pycache__
 .cache*
 *.pyc
 *.egg*
-docs/api/
 .ipynb*
+
+# Documentation Files
+docs/api/
+docs/_build/*
+docs/api.rst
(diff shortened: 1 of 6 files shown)


matthewkirby:
This is a fix for issue #104

matthewkirby:
I put >= version numbers into requirements.txt to resolve issue #106 as well.

PR #116, opened 2019-07-21
Issue/112/refactor framework

matthewkirby:
We removed a lot of code stubs and tests that are beyond the scope of the project. The existing code that we want to continue using was moved to `crypt` directories. We made empty files for the code currently being developed.

Once the following are checked, it will be ready for review. I will also assign reviewers then.
- [x] Get imports working correctly and remove the check functions


matthewkirby:
Imports are now working although I left in the check functions. I figure it is probably good to leave these in until we have some code in each file. I am afraid of docs crashing or something silly if we have no functions in files.

As far as I can tell, it is ready for review!

matthewkirby:
summary

Issue #120, opened 2019-07-22
Setup travis to install CCL and cluster_toolkit


matthewkirby:
This issue can be closed when:

- [ ] Travis runs a simple import test after installing CCL.
- [x] Travis runs cluster_toolkit

matthewkirby:
This is being pushed away from minimal example. For now, we are running with astropy until we are able to more reliably install CCL.

lbaumo:
I just realized I can use gcc 5 at slac, so I can let you know if it allows clusertoolkit to install

lbaumo:
okay, compiling with gcc 5 gives errors, so I would use 6 and higher to be safe

matthewkirby:
Great thanks for checking! At this point, gcc 6 or 7 or something is
installed. It’s just a matter of telling the cluster toolkit installer to
use the right install. Locally, I did this with os.environ[‘CC’]=‘gcc-9’
directly into the setup.py but we can’t do it that way on Travis.

aimalz:
Just leaving a comment to say #148 resolved the issue with cluster-toolkit.  I'm going to remove the priority tag for now because our tests have workarounds for CCL right now.

aimalz:
Now that CCL is pip/conda-installable, we can do a quick test to see if this issue can be quickly resolved.

combet:
Fixed in PR #299

PR #134, opened 2019-07-23
Issue/111/model definition

combet:
This refactors the model module to depend only on `cluster-toolkit`, `astropy`. Also works with `ccl` if it is installed. This comes with:
- tests
- documentation
- two example notebooks: `generate_mock_data` and `get_profile_model_values`


matthewkirby:
This looks good. Assuming tests run locally I am fine merging this in. I spent a bit tonight trying to set up travis locally to add cluster_toolkit but didn't manage to be successful.

aimalz:
This breaks Travis, but we need it to move on and will deal with that in #120.  It also needs some work within the scope of #9, but it's good enough for now.

Issue #149, opened 2019-08-01
Including shape noise in mock data can result in e1/e2 outside of physical domain


combet:
Currently, mock_data.py does not check that e1 and e2 are in [-1,1]. This can become problematic when adding a random gaussian shape noise to the mock data, if the requested shape noise dispersion is too large.

matthewkirby:
This was addressed and merged in PR #151. The PR addressed both issues 149 and 150 however this issue was never marked as resolved.

Issue #150, opened 2019-08-28
Swapping content of examples/support/mock_data.py to a function and updating to repo standards


matthewkirby:
Recently we discovered a small bug in our code that generates mock galaxy catalogs. The details are in Issue #149. While looking into this, I noticed that this file is not up to the same standards that the rest of our code is. I think this is a pretty big deal since this is a file that a potential user will use to learn about our code. So if this crashes/has bugs/is uncommented/is hard to read this is a major problem.

To close this issue
- [x] Change `MockData` to a function that returns a `Table` rather than a class that just contains a `Table`.
- [x] Make style adjustments 
- [x] Change the function to dynamically fill a `Table` along the way rather than having to keep track of many different lists of variables.
- [x] Revise `generate_mock_data.ipynb`
- [x] Revise `import_mock_data_as_GC_object.ipynb`
- [x] Revise `modeling_demo.ipynb`
- [x] Revise `run_function_on_GalaxyCluster.ipynb`
- [x] Resolve Issue #149 and merge it into this branch.
- [x] Merge the PR with all of these changes into master

While I am opening this as an issue for discussion/feedback/clarity, I am going to just hammer these changes out and open a PR ASAP. I have marked this a priority since it pertains to code currently within master.

PR #151, opened 2019-08-28
Issue/150/updating mock data.py

matthewkirby:
This PR revises `examples/support/mock_data.py` and updates the API calls in the example notebooks. This work was done in response to a bug discovered in this file where if `shapenoise` was set high enough, galaxies were generated that had aphysical values for `e1` and `e2`. While in there, I converted the class to a function (it was doing one thing only) and made it much more modular. Most of the documentation was rewritten as were variable and function names so please be sure to look at that during the review. 

When this PR is merged, we will be able to close Issue #149 and Issue #150 

I do have one outstanding question which I will copy from the Issue #150.
1. We set a mass definition for delta but we do not specify if this is with regard to the mean or critical overdensity. We should find this out and add this to the docstring in place of `XXX???`.

### DIFF
--- a/examples/generate_mock_data.ipynb
+++ b/examples/generate_mock_data.ipynb
@@ -4,7 +4,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## This example jupyer notebook uses clmm code to generate mock data with a set of configuration parameters, producing a catalog of ideal and noisy example data."
+    "## In this example we generate mock data with a variety of systematic effects including photometric redshifts, source galaxy distributions, and shape noise"
    ]
   },
   {
--- a/examples/generate_mock_data.ipynb
+++ b/examples/generate_mock_data.ipynb
@@ -10,9 +10,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "collapsed": true
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "import numpy as np\n",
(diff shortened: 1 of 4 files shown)


combet:
@matthewkirby, regarding the mean versus critical overdensity question. The old CLmassmod was working with critical, but we now use cluster-toolkit to provide all the WL functions and looking into it, it uses the mean overdensity convention.

matthewkirby:
In the future, when you make a review, you can click on lines in the code and add a comment. Other people can then comment directly on each point.

# Celine's Review
1. Good catched, I updated the docstrings to only have H0
2. Thanks for pointing this out. I changed it to `photoz_sigma_unscaled` since the actual photoz error is `photoz_sigma_unscaled (1+z)`
3. Agreed. The cell for making the cosmology object now has all 3 options available with astropy and ccl commented out.

I believe that addresses each of your comments @combet

matthewkirby:
# Partial Marina's review
## `mock_data.py`

1. "explicit docstring in function compute_photoz_pdfs" 

I am not entirely sure what you mean by this. I expanded on the docstring however to describe in more detail what the function is doing.

2. "suggestion : in config rename Delta as Delta_SO (for spherical overdensity)." 

This is a change to the API of some of the core functionality of the code. I believe that this is beyond the scope of this PR. My suggestion is to add this as a comment to Issue #152 

3. change mdef in generate_galaxy_catalog : if what we pass is the value of Delta this quantity should be called Delta (or Delta_SO) and not mdef. We can say in the docstring that Delta is computed wrt rho_mean and that the mass is then computed as a SO mass enclosed in radius Delta.

Good call. I made this change.

4. linked to the above comment and maybe for future development it would be great to be able to choose between critical or mean density for the mass definition.

I do feel that we should allow for this. However its good to keep in mind that this code is just for testing the main code. Anywhere we use a mass definition in the main code we should allow either wrt to critical or mean density. Lets also add this to issue so that we check for this before release #152 

5. the definition of the concentration parameter should be added. Is its definition homogeneous with that of the mass? This comment is more about modelling.py but it also applies here.

This should be in the same mass definition as the mass or it makes no sense. Within the scope of this PR, I have made mention of this in the docstring

6. In docstring change Choose the source galaxy distribution to be fixed or according to a predefined distribution. for Choose the source galaxy distribution to be fixed or drawn from a predefined distribution.

Made this small change to docstring.

7. suggestion :in generate_galaxy_catalog be more specific about the distribution called chang13, add explicit reference to the paper.

In my opinion, stating `Chang et al. 2013` should be sufficient in the docstring. If we start adding links to the code we have to worry that they remain valid. Regardless, I revised docstrings to include Chang et al. 2013 rather than chang13.

8. in generate_galaxy_catalog docsting : explained why we can have aphysical values for ellipticities (just copy paste what is written in _find_aphysical_galaxies)

I expanded on this a little in the docstring. Copy and pasting docstrings is just as bad as copy and pasting code! 

9. suggestion : in generate_galaxy_catalog be specific about the columns of the output table. It is : src_ra, src_dec, e1, e2, src_z, id when there is no photoz pdf created and src_ra, src_dec, e1, e2, src_z, pzbins, pzpdf, id when there are.

I did not make this change. As long as we return an astropy table, the table columns can be found by just `table.columns`. I mainly feel this way because this code is entirely for examples and not for development and science. The documentation also does not include these functions, these docstrings are entirely for our benefit.

10. Add reference and more docstring for the function draw_galaxy_positions. Why is there a 4 hard coded ? From which distribution are you drawing the galaxy position from ?

I have not addressed this point. I am not sure what you mean here. This function seems pretty straightforward to me. I am not sure what is hard coded. I expanded a little on the math in the docstring.

11. If possible, add reference for the shape noise and ellipticities computation in _generate_galaxy_catalog.

Shape noise is simply Gaussian here. If someone has a reference for the shear computations, please stick that in.

12. remove the commented line 262

I don't see any commented lines.

matthewkirby:
# Marina's Review, Part 2
### `mock_data.py`

13. in compute_photoz_pdfs : explicit that you model the scatter of the distribution as Gaussian and that you want the standard deviation of the distribution (ztrue - zphot)/(1+ztrue). Explicit also that the PDF of each zphot is computed as a gaussian of standard deviation equal to that of the (ztrue - zphot)/(1+ztrue) distribution.

I vastly revised the documentation in `generate_galaxy_catalog` to fully describe what we are doing. The docstrings of the internal functions can be minimal as they are internal functions it would only be for our benefit as developers. Everything should now be in the primary docstring.

14. suggestion : in compute_photoz_pdfs put the zbins arguments as input with default values to 0.5 and 0.03.

I made this function internal. We can add this enhancement if we want in the future but this seems like a small point at the moment.

matthewkirby:
# Marina's Review part 3 (final)
### `generate_mock_data.ipynb`
1. change the cosmology object import

 add explicit labels on the plots

 suggestion : add a section like Look at the data in the mock catalog after cell 9

 suggestion : change the catalog name for something more explicit than supermessy

 change comments all galaxies at z=0.3 for all galaxies at z = src_z (if I understand correctly here src_z=0.8 and not src_z=0.3).

matthewkirby:
# Marina's Review part 3 (final)
### `generate_mock_data.ipynb`

1. change the cosmology object import

This was done as part of Celine's review

2. add explicit labels on the plots

Done

3. suggestion : add a section like Look at the data in the mock catalog after cell 9

Done

4. suggestion : change the catalog name for something more explicit than supermessy

Done

5. change comments all galaxies at z=0.3 for all galaxies at z = src_z (if I understand correctly here src_z=0.8 and not src_z=0.3).

Yup, thanks for pointing this out. Corrected.

### `import_mock_data_as_GC_object.ipynb`
First, I renamed it.

1. change the cosmology object import

Done as per Celine's suggestion

2. suggestion : use the same name for the mock catalog as in generate_mock_data.ipynb.

Done

3. add explicit labels on the plots

Done

matthewkirby:
@combet @marina-ricci 

Please take a look. If I have addressed all of your review comments to your satisfaction, please accept the review. Otherwise, please let me know.

matthewkirby:
@marina-ricci 

Additionally, I revised the definition of e1 and e2 in the docstrings. I also revised the title in both jupyter notebooks (both the one you suggested adding and the other notebook).

> ### You addressed all the main points. `Mock_data.py` reads much better now !
> I have still some minor comments :
> 
> * [ ]  change the explanation of Delta_SO in `mock_data.py`

Great suggestion. I made a couple small edits but predominantly used your suggestion.
 
> * [ ]  There is a commented line l.153 `# galaxy_catalog['id'] = np.arange(ngals)`. Not sure it's here on purpose.

Removed it, thanks for pointing that out.

> * [ ]  Add reference for the shape noise and ellipticities computation. I am not 100% sure this is trivial.

1. Shape noise is straight up just Gaussian, we aren't doing anything special. We have a tangential shear as if the galaxy is circular, then we say that there is some amount of Gaussian noise on it due to the fact that it's actually elliptical. While there may be more complicated options, this is the simplest and I am not sure we would be able to find a reference.

2. Ellipticity. We have `e1 = -gamma_t cos(2phi)` and `e2 = -gamma_t sin(2phi)`. I took a look at some references and cannot reason this out quickly. I am going to ping Celine and Lucie in slack to try to work this one out.

> * [ ]   Change the explanation of shapenoise in  `mock_data.py`:
>   ` shapenoise : float, optional If set, applies Gaussian shape noise to the galaxy shapes`

Done.
 
> * [ ]  suggestion : add a small header in `import_mock_data_as_GalaxyCluster_object.ipynb` as in `generate_mock_data.ipynb`. Even if it's for internal purposes it's sometime good to remember the goals of the notebooks. 

Great point. These notebook aren't internal either! These are meant to be demos on how to use the code for a new user. We want these to be as clear and user friendly as possible!

> (small typo in the header of `generate_mock_data.ipynb` : `jupyer`)

Fixed, thanks.

marina-ricci:
Last point : there is a bug with the RA and Dec of the mock catalogs (because of a wrong unit conversion).
@combet corrected the code in her branch if I remember. 

- [ ] That need to be fixed before merging.

combet:
Thanks @marina-ricci for remembering the units problem! `get_angular_diameter_distance` returns `pc` but we need `Mpc` here. I corrected line 242 or `mock_data.py` accordingly. It should be okay now.

matthewkirby:
Great! Thanks for all the work everyone in getting this merged!

Issue #136, opened 2019-07-24
Small GalaxyCluster bug


matthewkirby:
To reproduce: 
```python
import clmm
from astropy.table import Table
x = clmm.GalaxyCluster('1', 161., 55, .3, Table())
```

This throws the error
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-5-0cf5e956986f> in <module>()
----> 1 x = clmm.GalaxyCluster('1', 161., 55, .3, Table())

~/anaconda3/lib/python3.6/site-packages/clmm-0.1-py3.6.egg/clmm/galaxycluster.py in __init__(self, unique_id, ra, dec, z, galcat)
     35             raise TypeError('ra incorrect type: %s'%type(ra))
     36         if not isinstance(dec, float):
---> 37             raise TypeError('dec incorrect type: %s'%type(dec))
     38         if not isinstance(galcat, Table):
     39             raise TypeError('galcat incorrect type: %s'%type(galcat))

TypeError: dec incorrect type: <class 'int'>
```

If a user passes an `int` ra or dec we should just convert to `float` and move on. 

To close this issue:

- [ ] Include a regression test in `test_galaxycluster.py`. Before touching `galaxycluster.py`, write a test that passes an `int` into both `ra` and `dec` and have them fail. Then after these tests are written, move onto step 2 below.
- [ ] Change behavior to allow `int` and convert to `float` and iterate until tests from step 1 pass.

matthewkirby:
Fixed by @m-aguena in PR 153.

PR #153, opened 2019-09-24
fix bug for int (and convertable str) passed to ra, dec, z

m-aguena:
to fix issue #136


matthewkirby:
Looks good! Thanks for fixing this!

Issue #157, opened 2019-10-22
Sign problem in gamma_x (cross shear) definition?


combet:
The cross shear is coded up as `gamma_x = - shear1 * np.sin(2.*phi) + shear2 * np.cos(2.*phi)`, but from g_t = -Re(g exp(-2i phi)) and g_x = -Im(g exp(-2i phi)), one gets reversed signs, i.e. `gamma_x = shear1 * np.sin(2.*phi) - shear2 * np.cos(2.*phi)`.

This should be double checked, confirmed and if so corrected. The tangential shear`gamma_t` appears to be coded OK. This happens in `polaraveraging.py` but should be checked for throughout the code.

Suggest doing so once PR 143 is processed.

shenmingfu:
The formulas `g_t = -Re(g exp(-2i phi))` and `g_x = -Im(g exp(-2i phi))` are correct. 
Reference: [Schneider, 2006] Schneider, P. (2006). Weak gravitational lensing. Gravitational lensing: strong, weak and micro, pages 269–451.
Arxiv: https://arxiv.org/abs/astro-ph/0509252
Then new formula `gamma_x = shear1 * np.sin(2.*phi) - shear2 * np.cos(2.*phi)` is correct.

shenmingfu:
The notebook I used for calculating `g_x` is `issue/114: demo_of_polaraveraging`, and it's using `import clmm.polaraveraging as pa`. In `pa`, the code uses `pa.compute_shear` and `pa.make_shear_profile` to get `g_x`.

shenmingfu:
`pa.compute_shear` uses `def _compute_cross_shear(shear1, shear2, phi):` to get g_x. https://github.com/LSSTDESC/CLMM/blob/b49f247158e9826d9e9f077edc6b7e524f47e2be/clmm/polaraveraging.py#L199

shenmingfu:
I realized the reason we had this sign issue is probably the definition of the direction of RA (and the definition of position angle phi) on the observation image and catalog. Now we use **RA from right to left**, but the position angle **phi is measured from x-axis (left to right) counter-clockwise**.

Issue #162, opened 2019-10-25
Make sure unit tests also validate with external values


m-aguena:
There are some (a lot) of unit tests of our functions that are only performing consistency checks (i. e. only checking results against themselves). We have to also check against known values.

m-aguena:
Has to be fixed in
- [ ]  `modeling.py`

aimalz:
Is this distinct from #158?

m-aguena:
Solved by PR #200

Issue #158, opened 2019-10-23
Validate physical functions with other libraries


m-aguena:
We should cross-validate the functions wth physical meaning used in `CLMM` with other libraries.

m-aguena:
Let's start with `NumCosmo`

m-aguena:
Adressed by PR #200

PR #168, opened 2019-11-18
fixed gx bug; corrected some typos

shenmingfu:
Check this for details.
https://github.com/LSSTDESC/CLMM/issues/157#issuecomment-555121787


tests/test_polaraveraging.py
```diff
@@ -9,25 +9,29 @@
...
-    cross_shear = pa._compute_cross_shear(shear1, shear2, phi)
-    testing.assert_allclose(cross_shear, expected_cross_shear)
-
-    shear1 = np.array([0.15, 0.40])
-    shear2 = np.array([0.08, 0.30])
-    phi = np.array([0.52, 1.23])
-    expected_cross_shear = [-0.08886301350787848, -0.48498333705834484]
-    cross_shear = pa._compute_cross_shear(shear1, shear2, phi)
-    testing.assert_allclose(cross_shear, expected_cross_shear)
+#    shear1, shear2, phi = 0.15, 0.08, 0.52
```
aimalz: We can remove these instead of just commenting them out now that we have new test cases implemented.

matthewkirby:
Also, I agree with Alex. We shouldn't have any commented out tests. If a test is commented it either 
A) is irrelevant (so just remove it)
B) doesn't work (so why doesnt it work? a bug? fix it!)

aimalz:
@matthewkirby My understanding is that we weren't consistently using the same sign convention between functions in our own code.  This just made all the functions in this module consistent with each other.

matthewkirby:
@aimalz Alrighty then. Is it possible to write a test that then compares the two pieces of code to ensure we are internally consistent?

Regardless, a validation test would still be excellent (and probably required for most of our science functions before a release). Is this something we can compute using NumCosmo (@pennalima @m-aguena) and drop into a txt file?

aimalz:
So the current tests in this branch do include a few null tests of self-consistency that weren't there before, i.e. the limiting cases of physics/geometry.  It would be entirely reasonable to have a test that pipes the output of the changed function through to another function and checks null tests (like sign consistency) for that.  However, I think that falls under #176 and shouldn't stand in the way of approving this PR.  Regarding replacing the magic numbers with the numerical outputs of similar functions in, say, `NumCosmo`, I requested that this be taken care of in #178, which would kill off both #158 and #162.

cavestruz:
Removed the commented out lines.  I think the comparison with other codes, next steps of validation tests, etc. is what @m-aguena is working on in PR  #178

matthewkirby:
Yup. I was going up from the bottom of the open PRs and had not yet seen @m-aguena. Totally agree that that falls under there.

matthewkirby:
My comment before was not a full review. For this new review, I did a complete review of all of the changed files.

aimalz:
@matthewkirby Is your request above about leaving the other commented tests as is?  I think those were the ones that were only recently commented out with immediate plans to revise them rather than to remove them.  My request was only about the comments in the test that had been changed, not the other commented lines in the test file.

cavestruz:
> Regarding the changes to `test_polaraveraging.py`
> 
>     * [ ]  In`test_compute_cross_shear()`, it looks like we removed quite a few tests. These tests should have been revised, not removed!
Looking at the diff with previous versions to see what we need to add back in.
> 
>     * Some tests were removed from `test_compute_tangential_shear()` as well but these were super pointless so no problem there.
> 
>     * [ ]  Many tests were removed from `test_compute_lensing_angles_flatsky()`. These should not have been removed. They were unrelated to this PR and are meant to be corrected in the open issue to revise the tests in this file. Please add these tests back in.
Looking at the diff with previous versions to see what we need to add back in. Adding these back in now.... 
> 
>     * [ ]  I noticed that an absolute tolerance is now being passed around. This is fine but we need to be very careful that this is defined appropriately. We don't want a test to pass because we have atol=1e-6 and the values are test=3e-7 and truth=5e-7. Unless atol has been included for a reason, I recommend removing it.

We added an atol because for the comparison, if the value in b is zero, then rtol doesn't impact anything.  We can decrease what atol is, perhaps to 1e-7.> 
>     * The changes to the docstrings look great too!

cavestruz:
If we're okay keeping the tests in test_compute_lensing_angles_flatsky() commented until a later PR (separate issue for this), then I think this is ready to merge.

matthewkirby:
> We added an atol because for the comparison, if the value in b is zero, then rtol doesn't impact anything. We can decrease what atol is, perhaps to 1e-7.>

Good catch, you are totally right. The number doesn't really matter, 1e-6 and 1e-7 are both fine.

You got the tests back too, great.

> If we're okay keeping the tests in test_compute_lensing_angles_flatsky() commented until a later PR (separate issue for this), then I think this is ready to merge.

Yep, that was my intent all along. I just didn't want to lose them =)

Approved!

matthewkirby:
Before you merge, make sure to merge master in on the command line (after pulling the one I am about to force merge) and manually deal with merge conflicts. There might be some in `polaraveraging.py`.

cavestruz:
OK - merged the latest master.  @aimalz - I think I need your approval before I (or whoever gets to it first) can squash and merge.

PR #143, opened 2019-07-24
Issue/114/rewrite shear azimuthl averager

marina-ricci:
Pool request to include into masters the code to construct shear profiles from data. This contains : 

- clmm/polaveraging.py
- test/polavering.py

Some notebooks

PR #169, opened 2019-11-18
Issue/77/lego pipeline

cavestruz:
@marina-ricci and I merged master into this branch and double checked that the demo_pipeline.ipynb still ran.


marina-ricci:
- Before merging, we should make sure to only merge the file we actually changed. There is a lot of old files here that should be removed by now (e.g. a pickle file with numbers).The only files that need to be merged are `demo_pipeline.ipynb` and `support/mock_data.py`.

aimalz:
@combet Can you try running this on NERSC?  @cavestruz and I encountered a problem activating the desc-clmassmod environment.

combet:
- Didn't succeed to run at NERSC. (same goes for #171 )
- But, running from my laptop, the demo won't run as it is as it's not updated yet to the new mock_data API. Checkout the `generate_mock_data` notebook to see how the call to generate mock data functions go.

matthewkirby:
Okay. Something funky is going on here. It might have to do with the weird git problems in a lot of these PRs from the Bochum workshop. Just in case, if you ever rename or remove a file that is tracked in a repository, make sure you use `git mv filename` or `git rm filename`. Otherwise, you can cause problems when things sync up. There are quite a few changes in this branch that are undoing things that I am 150% sure that I did in PR #143. I am going to open a review and inline comment then all.

matthewkirby:
Closing this PR. Please see PR #179 instead. This branch can be deleted once PR #179 has been merged.

PR #171, opened 2019-11-18
Issue/170/cleaning notebooks

marina-ricci:
This PR address issue 170. 
- I updated demo_of_polaraveraging.ipynb so that it works.
- I removed old files, namely, demo_polaraveraging.ipynb and demo_of_polaraveraging.py
- I resolved to bugs in clmm/plotting.py to make demo_of_polaraveraging.ipynb work.


aimalz:
As with #169, we should check that this runs on NERSC and then change the name of the demo notebook we're keeping to be consistent with the names of the other demo notebooks.

combet:
- No need to have a `config` dictionary anymore given that the new mock data functions take directly `cluster_m`, `cluster_z`, etc., as argument. 
- No need to specify `config['src_z'] = 0.8` (or src_z=0.8 when not using the dict) as the example uses the Chang redshift distribution.
- Add a `ax1.set_aspect('equal', 'datalim')` so that the distribution of (e1,e2) looks 'round'.
- The [`g_x` versus `g_t`] plot is not very informative; use histograms of `g_x` and `g_t` instead? `g_x` should then be centered on zero.
- Similarly to the `g_t` profile as a function of R[kpc] and R[Mpc], it would be good to have it as a function of angular distance in [deg].
- Last plot (cross shear profile) is missing units on the x-axis.

aimalz:
The remaining notebook still has all its outputs and needs to be cleared out.

matthewkirby:
I ended up making the style changes. Since I am the picky one on this anyways...

cavestruz:
Were folks able to run the notebook?  I'm getting an AttributeError:
image

matthewkirby:
I was able to run no problem. Did you reinstall clmm before launching jupyter? `python setup.py install --user`

combet:
I'm encountering the same problem as @cavestruz , although it was running fine when I checked it out this morning...

combet:
I just solved it by uninstalling and installing CLMM again.

cavestruz:
I did do the python setup.py install --user.  I'll try uninstalling and reinstalling again.

cavestruz:
Hi all - I believe I addressed all of the bullet points in the changes I made to the notebook.  This should be ready to merge.

matthewkirby:
@cavestruz comments look much better. Still several superfluous imports though. 
First,
- `import numpy as np` is never used (we can remove it)

Then,
- `import clmm.galaxycluster as gc` is never used
- `import clmm` IS used to load `GalaxyCluster`.
- Lets either make all of the clmm imports done via `import clmm` or by `import clmm.module as something` not a mixture of both. Another option would be to only import the thing you need. `from clmm.galaxycluster import GalaxyCluster` `from clmm.polaraveraging import compute_shear, make_bins, make_shear_profile` etc

cavestruz:
Okay, cool - picked the option to import specific thing.  I personally think this reads better for pedagogical purposes (i.e. easier to read and understand if we're onboarding someone).

matthewkirby:
@cavestruz Great! That is what we do in the code body as well so good choice for consistency! I am going to go ahead and approve my review but you should update this below. I didn't want to do it in case you were editing things and causing conflicts.

Change `from clmm import GalaxyCluster` to `from clmm.galaxycluster import GalaxyCluster` for consistency with the other imports.

cavestruz:
@matthewkirby - made the change.  If all is good, I'm going to squash and merge.

aimalz:
@combet The changes you requested have now been made.  Feel free to submit an approving review if/when you're satisfied.

combet:
Thanks for all the changes - it looks great! Although it got ticked above, this one little thing remains to be done. Then I'm happy for it to be merged to master.

- [ ] No need to specify src_z=0.8 as the example uses the Chang redshift distribution.

PR #174, opened 2019-11-19
Issue/110/re refactor clmm api

maho3:
This adds an io.py file which adds the load_from_dc2_rect function to load and save GalaxyCluster objects from GCR catalogs. Also added an example notebook demonstrating its use and added some type/value tests in test_io.py. 

Can't add output tests because Travis doesn't have access to GCRCatalogs.


matthewkirby:
Regarding the changes to `galaxycluster.py`, after I made the style changes and deleted the commented out line. It is no longer included as a changed file in the PR. So please disregard that first comment.

maho3:
I made some changes according to Matt's comments and marked which one's they address above. Some comments remain before I finish off the changes:

- I think the default ra/dec/z range makes sense as a demonstration for which ranges are appropriate. Though this isn't important if others want it removed.
- I think the for loop is inevitable, because we need to query the catalog in multiple ranges anyway. We could theoretically merge all the clusters info into one query for gcrcatalogs, but that would, I think, overcomplicate the function (lots of or's and and's). I think a more advanced loading function would be more appropriate to develop in a future version, if we need to reduce runtime.
- I made some helper functions to break the megafunction up. I still have to write tests for these.
- When you mention test the GCR functionality, do you mean in Travis tests or independently?

matthewkirby:
> I think the default ra/dec/z range makes sense as a demonstration for which ranges are appropriate. Though this isn't important if others want it removed.

Sounds good, lets keep them in then.

> I think the for loop is inevitable, because we need to query the catalog in multiple ranges anyway. We could theoretically merge all the clusters info into one query for gcrcatalogs, but that would, I think, overcomplicate the function (lots of or's and and's). I think a more advanced loading function would be more appropriate to develop in a future version, if we need to reduce runtime.

Alright then. Since I don't know how the gcr works I can't say anything else here.

> I made some helper functions to break the megafunction up. I still have to write tests for these.

This looks much more modular now! Awesome! See below for some new review comments for these functions.

> When you mention test the GCR functionality, do you mean in Travis tests or independently?

Independently. Unless we can access use it on travis (which would mean the reader and catalog would need to be public), we cannot test it on travis.

matthewkirby:
# `load_clusters_with_gcr.py`
##### `_get_filters_from_range`
- [ ] Docstring could use a rewrite. A little wordy. Suggestion
```
    """Create a filter for a given quantity in GCRCatalog format given a central value and range.

    Parameters
    ----------
    filter_name : str
        Quantity to filter
    center : float
        Central value of the filter
    filter_range: length-2 array-like of floats
        Range above and below the central value of the filter
    """
```
- [ ] It might be a little cleaner to just pass `low` and `high` (with better variable names) to the function and it makes the filter between those two ranges. It makes the function a little simpler and has all of the same functionality. You would just have to pass `('ra', center-ra_min, center+ra_max)` rather than `('ra', center, [ra_min, ra_max])`. As it stands, the documentation feels a little sketchy because of how the function is set up.
- [ ] If this is GCR specific, lets rename it as such. `_make_GCR_compatible_filter` or something. Also, be careful with `get` because it is a standard word in functions to mean you are just returning something that you already know. Here, we are creating the filters.

##### `_calc_reduced_shear`
- [ ] We have a function named this already (well, slightly different but almost the same) in a different module. Let's call this `_calc_reduced_shear_from_convergence` or something. 
- [ ] This is science! We want this in the main code body, not here. When you move it, it does not need to be a private function.
- [ ] Docstring

##### `load_from_dc2_rect`
- [ ] What does `rect` mean here? I see it in the docstring, we can probably just omit that part from the func name though.
- [ ] The way the function is written, we still cannot test it. My suggestion, this function is general. It takes a keyword arg with default like `_reader='GCR'` (not underscore for private).  If _reader=GCR, we call a different function that uses the GCR to return an astropy table or something of the results. We can then use _reader=something_else that calls a different function that returns an astropy table and we can do all of the other fun things that the code does here.
- [ ] The two checks for ra_range and dec_range don't make too much sense. These are modifiers for the central ra and dec of the cluster. If ra_range=(-350, 350) the test passes but the results will make no sense. We can probably not worry about this test and put a little bit of trust that the user is doing the right thing. When in doubt, we can add a paragraph to the docstring of this func explaining reasonable choices.
- [ ] This function is to use the GCR to read a catalog. Why are we doing any science? Lets axe the calls to `_calc_reduced_shear`

maho3:
Addressed almost all the comments except about _calc_reduced_shear. Waiting for the go-ahead to put this function in a 'science' module (e.g. modeling.py).

Made a spoof class, _test_catalog, to mimic GCRCatalogs which we can use for testing. Need to add more tests which take advantage of our new testing functionality.

matthewkirby:
Looks like you addressed everything that I had brought up, great! A couple of more suggestions

- [ ] For `_test_catalog`, it might be good to have a constructor that defaults to what you have set up but allows for arrays to be passed in if it helps testing (then you can have actual numbers to compare for outputs to `load_from_dc2`.
- [ ] Be careful with your imports. I am pretty sure that `from clmm import bleh` is not doing what you think it is doing. I expect that this is importing the installed version of clmm. To import code from our other modules, you want to do something like `from .galaxycluster import GalaxyCluster` and `from .modeling import thing`. Since you are a directory down, this exact syntax might not work, (maybe `from ..modeling import thing`?)
- [ ] Lets change string formatting to f-strings and be sure to run `pylint load_clusters_with_gcr.py` before we are done. For f-strings, this is as easy as changing `print('Loading cluster %s'%cl_id)` to `print(f'Loading cluster {cl_id}')` (Note the leading f).

matthewkirby:
@maho3 You should merge master in but otherwise looks good. I am not sure if anyone else wants to take a look first though

maho3:
Yeah, I'll do a last merge-master-in right before I ask someone to merge this into master. I'm stepping out for a seminar and after that I'll see if anyone has taken a look

cavestruz:
Some questions - load_from_DC2.ipynb is great for DC2 users and the overall LSST DESC community.  
- [ ] Should we link to instructions for access to NERSC/cori, and/or appropriate environments to use?  Picture a new DESC member who just got access to cori, so can log into cori, clone our repo, but that's it.

maho3:
Added Camille's requested links and merged master in

PR #177, opened 2019-11-19
Update requirements

aimalz:
I just added `cluster_toolkit` and `pyccl` to the requirements.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ install:
   - sudo apt-get install libgsl0-dev -y
 
   # Install the dependencies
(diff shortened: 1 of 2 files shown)


aimalz:
. . . And installing pyccl via our `setup.py` fails even though having the install command in `travis.yml` did not cause an error.  I'm probably making a really stupid mistake here.  @matthewkirby Can you point out where I've gone wrong?

aimalz:
Apparently I'm remembering this all wrong and cluster_toolkit is not pip-installable.  So, nevermind then!

PR #179, opened 2019-11-19
Issue/77/lego pipeline remastered

matthewkirby:
This PR is intended to replace PR #169 to avoid the funny business that was going on behind the scenes there.

Just mirroring the comments from that PR...

From @aimalz:
> @combet Can you try running this on NERSC? @cavestruz and I encountered a problem activating the desc-clmassmod environment.

From @combet 

> - Didn't succeed to run at NERSC. (same goes for #171 )
> - But, running from my laptop, the demo-pipeline notebook won't run as it is as it's not updated yet to the new mock_data API. Checkout the generate_mock_data notebook to see how the call to generate mock data functions go. Are you sure you reinstalled clmm before running the notebook? It shouldn't have run successfully. This call ideal_data = mock.MockData(config=config) is deprecated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ __pycache__
 *.pyc
 *.egg*
 .ipynb*
+*.pkl
 
 # Documentation Files
 docs/api/
--- a/examples/demo-pipeline.ipynb
+++ b/examples/demo-pipeline.ipynb
@@ -1,0 +1,404 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Fitting a WL mass using `clmm`\n",
+    "\n",
+    "_the LSST-DESC CLMM team_"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "from astropy.cosmology import FlatLambdaCDM\n",
+    "import clmm.polaraveraging as pa\n",
+    "import clmm.galaxycluster as gc\n",
+    "import clmm.modeling as modeling\n",
+    "import clmm\n",
+    "import sys\n",
+    "# This notebook must be run from the  \n",
+    "sys.path.append('./support')\n",
+    "import mock_data as mock\n",
+    "from numpy import random\n",
+    "from scipy import optimize as spo"
+   ]
+  },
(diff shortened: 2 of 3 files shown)


matthewkirby:
On top of copying the files over, I also

- Renamed the file to be consistent with other examples
- Added a line at the end to remove the pickle file being generated

matthewkirby:
Since I technically opened this PR, I cannot do a review. So review comments below.

- [ ] Run the linter and correct style
- [ ] Lets change from inline comments to markdown cells with comments for consistency with our other examples
- [ ] Why are we saving the cluster and then reopening it? Lets just load a cluster and use it. If we want to demo that feature, that should be done in the demo for `galaxycluster.py`.
- [ ] Move all imports to the top of the file.

Beyond the comments above (which are almost all just style and clarity), it looks great! Great job!

aimalz:
> Why are we saving the cluster and then reopening it? Lets just load a cluster and use it. If we want to demo that feature, that should be done in the demo for galaxycluster.py.
I think we're doing that to demonstrate the i/o syntax.  I think it's a good idea to keep it so the user can see how they'd modify the demo to, say, do something similar with their own data.

matthewkirby:
> @matthewkirby I think we're doing that to demonstrate the i/o syntax. I think it's a good idea to keep it so the user can see how they'd modify the demo to, say, do something similar with their own data.

Sounds good to me.

It sounds like the only TBD is ensure that it runs at nersc. But, if it runs locally, whether or not it runs at nersc should just be a question of is the nersc environment set up correctly. With that in mind, I am ready to merge this.

aimalz:
> Run the linter and correct style

I did this and corrected everything that didn't require renaming actual `clmm` functions/keywords.

matthewkirby:
I just went and ran the notebook and two more comments
- [ ] `%load_ext pycodestyle_magic` lets not ship it with this.
- [ ] Cell 19 is currently erroring out on me `TypeError: nfw_to_shear_profile() missing 1 required positional argument: 'gt_profile'`

matthewkirby:
It would also be good to make it consistent with the work @cavestruz has done in PR #171. Rather than `import clmm.polaraveraging as pa` do `from clmm.polaraveraging import compute_shear, make_bins, make_shear_profile` and the same for the other modules.

At the very least change `import clmm` to `import clmm.galaxycluster as gc` to be consistent with how imports are currently done from the other modules.

Issue #170, opened 2019-11-18
Cleaning notebooks in `CLMM/examples`


marina-ricci:
- Remove some old notebooks in `CLMM/examples`.
- Check that `demo_of_polaraveraging.ipynb` is working fine.

matthewkirby:
I just went through to ensure everything runs after my linting and wanted to comment here on my findings. There are a lot of notebooks in master.

- `run_function_on_GalaxyCluster.ipynb`: 
    - This isn't something we do anymore
    - This isn't runnable as it needs `core` and `GCData`. `core` is part of the old code base as is `GCData` (rumor has it this is staying? but idk I think it should be removed at this point.)
    - Matt Kirby Opinion: Remove it
- `rewrite_shearAzimuthalAverager.ipynb` 
    - This looks like it was made to develop code that was moved to a module.
    - The functions made here are all in master in python files.
    - Matt Kirby Opinion: This was a development file that shouldn't have made it to master, Remove it.
- `modeling_demo.ipynb`
    - Demos the functionality of `modeling.py`.
    - Looks great but not runnable!
    - API was updated to correct a redshift bug. This file was not updated to compensate.
    - Matt Kirby Opinion: Update this file to the current API and be sure that we test these things before merging a PR that edits any code in the `clmm/` directory.
- `import_mock_data_as_GalaxyCluster_object.ipynb`
    - What a name. Might want to think about renaming this maybe `generate_mock_GalaxyCluster.ipynb`?
    - Looks great and runs
    - Matt Kirby Opinion: Move code to `generate_mock_data.ipynb` and remove this notebook
- `generate_mock_data.ipynb`
    - Demos making mock data and playing with it
    - There is some code commented at the bottom using shearAzimuthalAverager. Let's remove that.
    - Why don't we show making a GalaxyCluster with this data here?
    - Matt Kirby Opinion: Add contents of `import_mock_data_as_GalaxyCluster_object.ipynb` here and remove `import_mock_data_as_GalaxyCluster_object.ipynb`
- `demo_pipeline.ipynb`
    - Looks great and runs.
    - Some of the plots have two units on the axis labels with a question mark. Lets figure out what units our functions are returning and correct that
    - Matt Kirby Opinion: Establish what units our functions are returning and why are they not Mpc h^-1? Add this to the function docstrings. Correct the axis labels in this notebook
- `demo_of_polaraveraging.py`
    - Works and looks great!
    - Matt Kirby Opinion: Fantastic!

[...]

PR #181, opened 2019-11-20
Linter all the things

matthewkirby:
Style changes for all of the code files. The test files also need some serious linting but since that is active development we can put that off until later (or take care of it on a case by case basis as they are revised).

I also changed all string formatting in the main code body to f-strings. They are a billion times clearer than all other string formatting options. If we have a variable, `name='matt'` we can format strings three different ways.
1. Old-style: `"my name is %s" % name`
2. New-style: `"my name is {}".format(name)`
3. f-strings (aka Newest-style): `f"my name is {name}"`


matthewkirby:
I revised the example notebooks to the updated APIs where appropriate. I had quite a few comments to make about each of the example notebooks. See the comment that I made in Issue #170. Should now be ready to be reviewed.

matthewkirby:
Note: This should be reviewed ASAP and merged into all development branches. The longer both of these steps take, the more merge conflicts are going to happen caused by tiny changes to the python files.

Issue #156, opened 2019-10-15
Consistent set of constants


matthewkirby:
We should aim at setting a consistent set of constants for our code. This involves writing a `constants.py` module and putting constants here. Rather than calling astropy's or CCL or some other code's CLIGHT_KMS, we should have a `clmm.CLIGHT_KMS` that we always call instead. In `constants.py`, we can then choose whether to set our value equal to astropy/CCL/anything else. 

While speed of light is a silly example since most sources agree, things like newton's constant G, or the critical density rho_c (in a fiducial cosmology) have more deviations. For a demonstration, see this work that I did for CCL ages ago, https://github.com/LSSTDESC/CCL/issues/78. Many sources report the values to various degrees of precision and on top of that, report different values. At the end of the day, we should aim to pull these constants from CCL. If we have a `constants.py` and we always use our personally defined constants, updating every G value to a slightly different value only needs to be changed in one place.

Some examples of numbers that we may wish to implement this for can be drawn from CCL's choices. https://github.com/LSSTDESC/CCL/blob/master/src/ccl_core.c

- [x] Make `constants.py`
- [x] Convert our code to only use our definitions.

m-aguena:
We have to be careful with the numbers we put in this file, including the number of decimal cases and the source of the values. The uncertainty on those numbers determines the lower boundary for the uncertainties of the whole code!

I Reopened this issue so the people who added the values to take another look at the numbers and include sources.

cavestruz:
- [ ] References look like they were already added in constants.py (e.g. astropy 2014, and pdg 2013, and GSL - same reference sources as GSL, except for GNewton that needed tweaking, etc.)

PR #185, opened 2019-11-20
Issue/156/consistent constants

matthewkirby:
Following issue #156, I have made a constants file and included all constants that we use so far in the code. I have set it up as a special type of class called an enumeration along with documentation.

This file does not need to be tested so coverage will drop here.

### DIFF
--- a/clmm/constants.py
+++ b/clmm/constants.py
@@ -1,0 +1,15 @@
+""" Provide a consistent set of constants to use through CLMM """
+import astropy.constants as astropyconst
+import astropy.units as u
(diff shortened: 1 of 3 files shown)


matthewkirby:
With this merged, PR #156 can be closed and one more step can be crossed off of the v1.0 milestone!

matthewkirby:
Incidentally, the tests are counting these lines as tested as the unit tests are hitting the functions that access the constants. Granted, that value is not being tested, but thats another story for another day.

PR #178, opened 2019-11-19
Issue/158/validate physical functions

m-aguena:
Added notebook comparing CLMM output with `NumCosmo`. It can also run with some saved values computed by `NumCosmo` without having to actually run `NumCosmo`.

Issue #173, opened 2019-11-18
Add redshift information to output of make_shear_profile


m-aguena:
Make `polaraveraging.make_shear_profile` also return a mean redshift per radial bin

combet:
This has been done and merged. Closing the issue.

PR #188, opened 2019-11-20
Issue/162/unit tests with physical values

m-aguena:
added `NumCosmo` values to unit tests of modelling


m-aguena:
@cavestruz @matthewkirby @aimalz Right now the unit tests of modeling are failing because the precision of `get_critical_surface_density` in Travis is 2e-4 comparing to the value I got with `NumCosmo`. However when I compare with the value I get in my pc and in the binder @vitenti made I have a precision of 1e-8. I am using the same astropy version (v3.2.3) as Travis so I don't know where this difference came from. I don't know the best approach to proceed, do you think this is something to be investigated or just decrease the precision used in the unit tests?

matthewkirby:
Sounds like a bug to me. We should track it down at least why it’s
different. That function of ours you listed isn’t actually tested in or
unit tests. So it may be wrong. Check the values of the speed of light and
Newton’s constant between the versions as well because that could also be
the source.

matthewkirby:
When this is ready for a review, tag it up as such. I want to note that my comment from your other open PR applies here. It is really really bad practice to use a pickle file like this. You can't see inside and there is no limitation to what that code could do. For instance, it could install a bunch of dependencies and then malware and you can't know for sure until you open it and try. There really is no benefit to saving the numcosmo results in a pickle file over a text file so lets just use a text file.

cavestruz:
Quick question -  I noticed there's an rtol defined but no atol @m-aguena .  When Shenming and I were looking at the test_polaraveraging.py functions, both atol and rtol needed to be non-zero to account for certain cases.  I haven't carefully looked at the code yet, but this could be leading to the/a bug.  Happy to jump in on this first thing tomorrow and pair-code if that is helpful.

m-aguena:
> When this is ready for a review, tag it up as such. I want to note that my comment from your other open PR applies here. It is really really bad practice to use a pickle file like this. You can't see inside and there is no limitation to what that code could do. For instance, it could install a bunch of dependencies and then malware and you can't know for sure until you open it and try. There really is no benefit to saving the numcosmo results in a pickle file over a text file so lets just use a text file.

I just removed the picked file and put all saved values to txt file

m-aguena:
I added some tests where the computation of sigma_crit was done explicitly in the test function and got the same result. Than I noticed that the constants in master changed. They should have the same values but somehow because they were defined in different units and than converted there might be a difference there. I will investigate but it was either
```
# old
clight_pc_s = constants.c.to(units.pc/units.s).value
# new
CLIGHT_KMS = astropyconst.c.to(u.km/u.s).value
clight_pc_s = const.CLIGHT_KMS.value * 1000. / const.PC_TO_METER.value
 ```
or
```
# old
gnewt_pc3_msun_s2 = constants.G.to(units.pc**3/units.M_sun/units.s**2).value
# new
GNEWT = astropyconst.G.to(u.m**3/u.kg/u.s**2).value
gnewt_pc3_msun_s2 = const.GNEWT.value * const.SOLAR_MASS.value / const.PC_TO_METER.value**3
```

**Comment by m-aguena** [comment]
Finally found the problem, the `SOALR_MASS` constant didn't have enough decimal places. I updated it from `1.98892e30` to `1.9884754153381438e+30` and the tests passed. A good example of the importance of test with actual physical values.
@cavestruz @aimalz @matthewkirby

**Comment by m-aguena** [comment]
I reverted the value of solar mass to what was defined before based on ccl and added a conversion in the unit tests to correct for the difference in the physical values.

**Comment by cavestruz** [comment]
@m-aguena - I just received a request for review.  Just checking, is this "ready for review" (in which case, can you tag "PR: ready for review") or was I just added for the future when it is ready (in which case, can you tag "PR: not ready for review")?  Thanks for pushing this forward!

**Comment by m-aguena** [comment]
> @m-aguena - I just received a request for review. Just checking, is this "ready for review" (in which case, can you tag "PR: ready for review") or was I just added for the future when it is ready (in which case, can you tag "PR: not ready for review")? Thanks for pushing this forward!

I didn't know we were using that tag. Added it now

File: tests/test_modeling_physical_validation.py [hunk 1]
```diff
@@ -0,0 +1,161 @@
...
+import ast
+import astropy
+from numpy import testing as tst
+import numpy as np
+import clmm
+from clmm.constants import Constants as clmmconst
+
+# Read test case
+CODE_PATH = '/'.join(os.path.abspath(__file__).split('/')[:-2])
+VAL_FILES_PATH = '%s/tests/physical_values_nc'%CODE_PATH
```
matthewkirby: We shouldn't need to add anything to path so relative path work. `pytest` runs from the root package directory.

tests/test_modeling_physical_validation.py
```diff
@@ -0,0 +1,161 @@
...
+with open('%s/float_vals.txt'%VAL_FILES_PATH, 'r') as f:
+    TEST_CASE = ast.literal_eval(f.read())
+f.close()
```
matthewkirby: This shouldn't be global. Lets put this into a function that our tests can call.

tests/test_modeling_physical_validation.py
```diff
@@ -0,0 +1,161 @@
...
+NC_PROF = np.genfromtxt('%s/numcosmo_profiles.txt'%VAL_FILES_PATH,
+                        names=True)
+NC_DIST = np.genfromtxt('%s/numcosmo_angular_diameter_distance.txt'%
+                        VAL_FILES_PATH, names=True)
+
+R3D = np.array(NC_PROF['r3d'])
```
matthewkirby: Same as last comment. Lets put this into a function (along with everything that reads in from files).

tests/test_modeling_physical_validation.py
```diff
@@ -0,0 +1,161 @@
...
+TEST_CASE_SIGMAC_PHYSICAL_CONSTANT = (TEST_CASE['lightspeed[km/s]']*1000./TEST_CASE['pc_to_m'])**2/\
+    (TEST_CASE['G[m3/km.s2]']*TEST_CASE['Msun[kg]']/TEST_CASE['pc_to_m']**3)
+SIGMAC_PHYSCONST_CORRECTION = TEST_CASE_SIGMAC_PHYSICAL_CONSTANT/CLMM_SIGMAC_PHYSICAL_CONSTANT
+
+# Cosmology
+#COSMO_APY = astropy.cosmology.core.FlatLambdaCDM(H0=70, Om0=0.27, Ob0=0.045)
+COSMO_APY = astropy.cosmology.core.FlatLambdaCDM(H0=TEST_CASE['cosmo_H0'],
+                                                 Om0=TEST_CASE['cosmo_Om0'],
+                                                 Ob0=TEST_CASE['cosmo_Ob0'])
+COSMO_CCL = clmm.cclify_astropy_cosmo(COSMO_APY)
```
matthewkirby: Lets move this into functions

tests/test_modeling_physical_validation.py
```diff
@@ -0,0 +1,161 @@
...
+GAMMA_PARAMS = {
+    'mdelta':TEST_CASE['cluster_mass'],
+    'cdelta':TEST_CASE['cluster_concentration'],
+    'z_cluster':TEST_CASE['z_cluster'],
+    'z_source':TEST_CASE['z_source'],
+    'cosmo':COSMO_CCL,
+    'delta_mdef':TEST_CASE['mass_Delta'],
+    'halo_profile_model':TEST_CASE['density_profile_parametrization'],
+    'z_src_model':'single_plane',
+    }
```
matthewkirby: Move into function. We can just write one function that returns all of this stuff that each test calls.

tests/test_modeling_physical_validation.py
```diff
@@ -0,0 +1,161 @@
...
+
+def test_physical_constants():
+    '''
+    Test physical values of physical_constants
+
+    Notes
+    -----
+        The precision set for these tests are somewhat arbitrary
+        has to be better defined at some point
```
matthewkirby: This is not correct. numbers are directly from CCL. The only thing that isn't 100% identical to CCL is GNEWT. Once we have CCL integration, we will pull values from there.

tests/test_modeling_physical_validation.py
```diff
@@ -0,0 +1,161 @@
...
+def test_get_da():
+    '''
+    Test physical values of Da
+    '''
+    dl_clmm = clmm.modeling.get_angular_diameter_distance_a(COSMO_CCL,
+                                                            TEST_CASE['aexp_cluster'])
+    ds_clmm = clmm.modeling.get_angular_diameter_distance_a(COSMO_CCL,
+                                                            TEST_CASE['aexp_source'])
+    dsl_clmm = clmm.modeling.get_angular_diameter_distance_a(COSMO_CCL,
+                                                             TEST_CASE['aexp_source'],
```
matthewkirby: When the import scheme is changed above (`from clmm.modeling import get_angular_diameter_distance_a, otherfunc2, otherfunc3`) this line will be much shorter and we should be able to get the assert statements on the same lines as the computations.

matthewkirby: We want this because if a test fails, we want the error message to include the function call that failed, not just `dl_clmm`

tests/test_modeling_physical_validation.py
```diff
@@ -0,0 +1,161 @@
...
+if __name__ == '__main__':
+    test_physical_constants()
+    test_rho()
+    test_sigma()
+    test_get_da()
+    test_sigmac()
+    test_delta_sigma()
+    test_gammat()
+    test_kappa()
+    test_gt()
```
matthewkirby: Shouldn't have a main funciton like this.

tests/test_modeling_physical_validation.py
```diff
@@ -0,0 +1,161 @@
...
+Tests for modeling
+"""
+
+import os
+import ast
+import astropy
+from numpy import testing as tst
+import numpy as np
+import clmm
+from clmm.constants import Constants as clmmconst
```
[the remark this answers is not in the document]
m-aguena: fixed

tests/test_modeling_physical_validation.py
m-aguena: fixed

tests/test_modeling_physical_validation.py
m-aguena: fixed

tests/test_modeling_physical_validation.py
m-aguena: fixed

tests/test_modeling_physical_validation.py
m-aguena: fixed

tests/test_modeling_physical_validation.py
m-aguena: fixed

tests/test_modeling_physical_validation.py
m-aguena: What I meant is the precision I set there were somewhat arbitrary.

tests/test_modeling_physical_validation.py
m-aguena: fixed

tests/test_modeling_physical_validation.py
m-aguena: fixed

m-aguena:
> A bunch of things.
> 
> 1. There are a lot of small changes that are beyond the scope of this PR. the single line changes from `modeling.py` and `test_polaraveraging.py` for instance. They are done so let's just leave them.
> 
> * [x]  Move the numcosmo data files into the `tests/data` directory. We can either make a directory `tests/data/numcosmo/`, put them in there, and change names to `angular_diam...etc`   or just put them in `tests/data/` with their current names `numcosmo_angular...etc`.
> * [x]  Recover `tests/test_modeling.py`. This one is important. If you delete the file like this, when I go to merge the refactored tests, it will just delete the file. Just recover it and leave it untouched. Once this is merged into master, I will manually put the contents of `test_modeling_physical_validation.py` into the appropriate places in `test_modeling.py` (before merging my PR that writes these tests).
> * [x]  Lets be a bit more descriptive with the names of the numcosmo files. `float_vals.txt` for instance. Maybe this can be `numcosmo_config.txt` or something?
> * [x]  See inline comments for `test_modeling_physical_validation.py`. Generally, all of the code should be in a function. We also shouldn't have a main function in the test files. If you ever want to run a specific test only you can run `pytest path/to/test/file.py`.
> * [x]  This is nitpicky but docstrings should be 3x double quotes `"""` not single quotes `'''`. Lets just find and replace this before merging it in to be consistent across the project.
> * [x]  We should aim to have each assert statement have the function call within it rather than saving to a variable and then asserting.
> * [ ]  Lets also make a global variable `TOLERANCE` that has an rtol and atol value used for all of the tests rather than seemingly pulling different tolerances out of a hat for each test. This should be consistent with the rest of the test files.
> * [ ]  All notebooks in the examples directory should be examples that anyone who comes along to use the code can run. This is not true for this new notebook. Also, it isn't an example. If we have replaced the contents of this with automated validation tests, we no longer need this notebook.

@matthewkirby 

About the `TOLERANCE`, how should we deal if I know different components will have different agreement levels with NumCosmo? Should we level by the lowest tolerance?

About the notebook, I agree it does not have the same functionality as the other ones, and the comparisons in there were passed to the unit tests in `tests/test_modeling_physical_validation.py`. However I think it is very useful as it provides a guideline to compute the same properties with CLMM and NumCosmo. It probably has to be renamed and maybe moved somewhere else? (where I don't know though). Anyhow I think doing this should be in the scope of Issue #178

tests/test_modeling_physical_validation.py
matthewkirby: Ahhh gotcha.

matthewkirby:
> About the TOLERANCE, how should we deal if I know different components will have different agreement levels with NumCosmo? Should we level by the lowest tolerance?

Good point, these tests are not the same as the other. Doesn't make as much sense here.

> About the notebook, I agree it does not have the same functionality as the other ones, and the comparisons in there were passed to the unit tests in tests/test_modeling_physical_validation.py. However I think it is very useful as it provides a guideline to compute the same properties with CLMM and NumCosmo. It probably has to be renamed and maybe moved somewhere else? (where I don't know though). Anyhow I think doing this should be in the scope of Issue #178

Alright, we can figure that out there then.

matthewkirby:
Assuming no one else has anything to say, I am happy merging this.

m-aguena:
@matthewkirby I looks like everything is still working after merging master that had #173 in it. Would you like to merge it so that you can move on with #200 ?

matthewkirby:
Lets do it!

Issue #193, opened 2019-11-21
The YOLO test in `utils.py`


matthewkirby:
Sometimes, tests on travis crash. Never locally, only ever on travis. The test that crashes is this one from `test_utils.py` in the function `test_compute_radial_averages()`

```python
# Test a much larger, random sample with unevenly spaced bins
binvals = 10.0 * np.random.random(100)
xbins2 = [0.0, 10.0 * np.random.random(), 10.0]
inbin1 = binvals[(binvals > xbins2[0]) & (binvals < xbins2[1])]
inbin2 = binvals[(binvals > xbins2[1]) & (binvals < xbins2[2])]
assert_allclose(compute_radial_averages(binvals, binvals, xbins2, error_model='std/n'),
                [[np.mean(inbin1), np.mean(inbin2)], [np.mean(inbin1), np.mean(inbin2)],
                 [np.std(inbin1)/len(inbin1), np.std(inbin2)/len(inbin2)]], **TOLERANCE)
assert_allclose(compute_radial_averages(binvals, binvals, xbins2, error_model='std'),
                [[np.mean(inbin1), np.mean(inbin2)], [np.mean(inbin1), np.mean(inbin2)],
                 [np.std(inbin1), np.std(inbin2)]], **TOLERANCE)
```

The output that travis sees is
```
x and y nan location mismatch:
x: array([[4.834679,      nan],
               [4.834679,      nan],
               [2.904419, 0.      ]])
y: array([[4.834679,      nan],
               [4.834679,      nan],
               [2.904419,      nan]])
```

None of these elements should be nan so something is going wrong. I am going to take a look at this (so we can fix the sometimes crashes).

My intent with this test was to 
1. Generate 100 random numbers between 0 and 10
2. Generate one random number between 0 and 10 and then make bin edges [0, random, 10]. This will give us two bins with a random width.

My expectation is that the nans are coming when the bins are something like [0,9.8,10] and then there are no random numbers drawn between [9.8,10]. This seems unlikely as it keeps coming up. I may have a typo in my intents as well. Maybe a fix is to increase the number of objects in binvals, less likely to have zero objects in a bin? Or we could write a big list of random numbers (100, or 1000 or something and put the file in `tests/data/filename.txt`), read that in, say the bins are like [0.0, 6.3, 10.] or something and just manually compute it and test this rather than doing it randomly.

matthewkirby:
[...]

PR #194, opened 2019-11-21
Fix the YOLO test

matthewkirby:
Fixed the test that randomly crashed on travis as per issue #193


matthewkirby:
Before: **Randomly generate 100 points** and **two bins** between 0 and 10. Compute expected statistics and compare to the output of our function.

Now: **Load 1000 points from file**. Compute truth in **three** bins and compare to the output of our function.

This is because travis uses parallel threads/processes (idk which) and a test using randoms like this is not thread-safe. This is why it locally works when you run it (because 1 process) and not on travis.

Issue #189, opened 2019-11-21
User Feedback on Documentation


cavestruz:
- [x] Add plots to example code in readthedocs (needs to come from ipynb with outputs)
- [x] Consistency in Notebook naming conventions
- [x] "Auto-compile" the ipynb so we have the .rst files for readthedocs.io
- [x] After ^, remove the lines in CONTRIBUTING.md that talk through individually converting ipynb to rst

cavestruz:
Feedback from Peter Brown @pbrown801
- [ ] git clone -- need https://github.com/LSST/DESC/CLMM.git
- [ ] need to do python setup.py install --user
- [x] fails with syntax error - possibly a python version, saw difference between `which python` and `python -V` Definitely need python 3.  Need to make clearer in doc that python 3 is needed. 
- [ ] Consider moving requirements above installation.

shenmingfu:
> Feedback from Peter Brown @pbrown801
> 
> * [ ]  git clone -- need https://github.com/LSSTDESC/CLMM.git
> * [ ]  need to do python setup.py install --user
> * [x]  fails with syntax error - possibly a python version, saw difference between `which python` and `python -V` Definitely need python 3.  Need to make clearer in doc that python 3 is needed.
> * [ ]  Consider moving requirements above installation.

I will open another issue for these. Regarding "Consider moving requirements above installation", I guess I will also copy those newly added instructions in #191.

aimalz:
- [ ] When running `pytest` after installing `clmm`, we don't check that the user is doing so with Python 3.7.  Is there a way to do this?

shenmingfu:
The "example" link on `README.md` is currently linked to an old branch. But issue/196 is closed. I will change it to master branch and make a commit in issue/189. @matthewkirby

shenmingfu:
There are still some minor problems in README.md (difference between '_' and '-'; missing '_' or '-').  They don't strongly affect reader/user to understand the readme. We can fix them in a future issue.

matthewkirby:
Yeah. Lets not worry about minor things like that. Before v1.0 we should go through and scour spelling, grammar, small typos and fix em all. Might as well wait until we are at that stage though rather than doing it multiple times.

matthewkirby:
> When running pytest after installing clmm, we don't check that the user is doing so with Python 3.7. Is there a way to do this?

[...]

PR #199, opened 2019-11-22
Enforce minimum python version

matthewkirby:
Enforcing the minimum python version in the setup.py

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,12 @@ language: python
 cache: pip
 
 python:
+- "2.7"
+- "3.4"
+- "3.5"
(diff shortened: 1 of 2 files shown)


matthewkirby:
I ran travis with a boatload of python versions...
image

- 2.7 and 3.4 have a red ! because astropy v3.0+ doesn't support these
- 3.5 doesn't support f-strings

It looks like the minimum is 3.6

matthewkirby:
@aimalz Addressing your comment in Issue #189 about enforcing a python version. Minimum version right now is 3.6. We now enforce this.

Issue #201, opened 2019-11-23
quick fix of testing files


maho3:
- [x]  test test_load_clusters_with_gcr.py should not leave residual `.p` files after testing

- [x]  Remove 'io.py' from test_load_clusters_with_gcr.py docstring

- [x] remove test_travis.py because it's breaking the docs

maho3:
Closed by #202

PR #202, opened 2019-11-23
Issue/201/quick fix of testing functions

maho3:
quik fix to address #201


matthewkirby:
Taking a look now. I want to compile it on rtd and see if it fixed the problem and run the tests locally to ensure the fix worked before merging. For some reason I can't authenticate on the amtrak wifi...

PR #203, opened 2019-11-23
Issue/173/add z mean in make shear profile output

m-aguena:
Add z mean in radial bins to output of `make_shear_profile`. Apparently `modelling` functions `predict_tangential_shear`, `predict_convergence` and `predict_reduced_tangential_shear` also work if you pass an array to `z_source` with the same size of `r_proj`, so they don't have to change.

### DIFF
--- a/clmm/polaraveraging.py
+++ b/clmm/polaraveraging.py
@@ -273,10 +273,13 @@ def make_shear_profile(cluster, angsep_units, bin_units, bins=10, cosmo=None,
                                                     xbins=bins, error_model='std/n')
     r_avg, gx_avg, gx_err = compute_radial_averages(source_seps, cluster.galcat['gx'].data,
                                                     xbins=bins, error_model='std/n')
+    r_avg, z_avg, z_err = compute_radial_averages(source_seps, cluster.galcat['z'].data,
+                                                    xbins=bins, error_model='std/n')
 
-    profile_table = Table([bins[:-1], r_avg, bins[1:], gt_avg, gt_err, gx_avg, gx_err],
+    profile_table = Table([bins[:-1], r_avg, bins[1:], gt_avg, gt_err, gx_avg, gx_err,
+                           z_avg, z_err],
                           names=('radius_min', 'radius', 'radius_max', 'gt', 'gt_err',
-                                 'gx', 'gx_err'))
+                                 'gx', 'gx_err', 'z_avg', 'z_err',))
 
     if add_to_cluster:
         cluster.profile = profile_table
--- a/clmm/polaraveraging.py
+++ b/clmm/polaraveraging.py
@@ -279,7 +279,7 @@ def make_shear_profile(cluster, angsep_units, bin_units, bins=10, cosmo=None,
     profile_table = Table([bins[:-1], r_avg, bins[1:], gt_avg, gt_err, gx_avg, gx_err,
                            z_avg, z_err],
                           names=('radius_min', 'radius', 'radius_max', 'gt', 'gt_err',
-                                 'gx', 'gx_err', 'z_avg', 'z_err',))
+                                 'gx', 'gx_err', 'z', 'z_err',))
 
     if add_to_cluster:
         cluster.profile = profile_table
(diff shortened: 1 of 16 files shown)


m-aguena:
> Thanks @m-aguena, this looks good!
> 
> I'd just suggest adding a cell in `demo_polaraveraging_functionality.ipynb` so the user can see what the profile table looks like and that it now includes the average redshift and corresponding error. Simply the first few lines, like `cl.profile[0:4]`.
> 
> I will also open a separate issue to make an example notebook doing the full pipeline, but using non-ideal mock data, with a distribution of redshift for the background sources. If all goes well, the fit should work straightforwardly!

I added plots for the redshift of the sources. This was a great suggestion as, by implementing it, I found that sometimes there are `z_src<z_cl` (look at the `z_src` distribution plot in cell 6). @combet Is this something we should worry about at this point?

combet:
Hmm, thanks for pointing the redshift issue out @m-aguena. Nice catch! For the purpose on this PR, it's not something to worry about. But it is an issue... The mock data redshift generator should truncate the Chang et al. (2013) distribution at `zsrc_min = cluster_z + 0.1` (see line 187 of `mock_data.py`). So this should not happen, but it's clearly happening for a few sources (see plot below, where the red line corresponds to `cluster_z + 0.1`). So it's a bug and needs to be dealt with in a separate issue.
image.

clmm/polaraveraging.py
```diff
@@ -256,6 +257,8 @@ def make_shear_profile(cluster, angsep_units, bin_units, bins=10, cosmo=None,
     if not all([t_ in cluster.galcat.columns for t_ in ('gt', 'gx', 'theta')]):
         raise TypeError('Shear information is missing in galaxy catalog must have tangential' +\
                         'and cross shears (gt,gx). Run compute_shear first!')
+    if 'z' not in cluster.galcat.columns:
+        raise TypeError('Missing galaxy redshifs!')
```
matthewkirby: typo in redshifts

tests/test_polaraveraging.py
```diff
@@ -1,15 +1,13 @@
...
-import os
-import pytest
 
+import clmm
+import clmm.polaraveraging as pa
 
-rtol = 1.e-7
-atol = 1.e-7
+RTOL = 1.e-7
+ATOL = 1.e-7
```
matthewkirby: Lets convert this to `TOLERANCE={'atol':1.e-7, 'rtol':1.e-7}` and then we can just `**TOLERANCE` into the asserts.

tests/test_polaraveraging.py
```diff
@@ -155,82 +160,104 @@ def test_compute_shear():
...
+    testing.assert_allclose(profile2['gx'], expected_cross_shear[:-1], rtol=RTOL, atol=ATOL,
                             err_msg="Cross shear in bin not expected")
 
 
+if __name__ == '__main__':
+    test_compute_cross_shear()
+    test_compute_tangential_shear()
+    test_compute_lensing_angles_flatsky()
+    test_compute_shear()
+    test_make_shear_profiles()
```
matthewkirby: No main function

tests/test_polaraveraging.py
m-aguena: done

tests/test_polaraveraging.py
m-aguena: fixed

clmm/polaraveraging.py
m-aguena: fixed

m-aguena: fixed

m-aguena:
> Honestly, it is really hard to see what you changed in these files because of the style changes. We should avoid making style changes to lines that are not new to the PR because of this.
> 
> * [ ]  I don't see a unit test that checks the output value of z and zerr. We should certainly have this.

unit tests added

matthewkirby:
I am happy with this PR as well  now! Great job!

Lets be careful when we merge these two PRs by Michel. It might be best if you do it @m-aguena. Merge one of them with the button. THen pull master, merge master into your local branch and make sure everything is good. Then Do the second.

m-aguena:
> I am happy with this PR as well now! Great job!
> 
> Lets be careful when we merge these two PRs by Michel. It might be best if you do it @m-aguena. Merge one of them with the button. THen pull master, merge master into your local branch and make sure everything is good. Then Do the second.

@matthewkirby Ok, I will merge this first as it involves more people, than I can pull master into #162.

Issue #209, opened 2019-12-03
CLMM version in __init__.py and in all notebooks


cristobal-sifon:
- [x] move version definition from ``setup.py`` to ``__init__.py`` and read from there, as in [``pip``](https://github.com/pypa/pip/blob/master/setup.py). This will create a ``__version__`` attribute.
- [x] print ``CLMM.__version__`` in all notebooks

PR #210, opened 2019-12-03
Issue/209/version

cristobal-sifon:
This PR addresses both points raised in #209


cristobal-sifon:
will do. Does it make sense though rather than duplicating it, to have ``get_version`` in, say, ``__init__.py``? Any other suggestions?

matthewkirby:
There must be a standard because I expect that others run into this problem.

matthewkirby:
Actually. For docs to compile correctly, the package needs to be installed. Its possible that we can just `import clmm` and then say `clmm.__version__` in the docs config.

matthewkirby:
I just tested this and we can do something like this
```
from clmm import __version__ as clmm_version
version = clmm_version
release = version
```

cristobal-sifon:
OK let me try that

matthewkirby:
Looks great! Merging!

PR #232, opened 2020-02-11
Issue/231/Update bleeding edge to nightly.

matthewkirby:
We were running travis on 3.8-dev to test against the newest changes. Changed to nightly to avoid having to update this in the future. 

In general, failures on the `-dev` version and now `nightly` version can probably be safely ignored and put on the back-burner. No one using the code will be using these versions but they reflect potential future problems. To do this, I have also added the `allowed_failure` section to the config file for the `nightly` build.  If the `nightly` build fails we will see it but the overall travis build will pass (assuming `3.6` succeeds).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ cache: pip
 
 python:
 - "3.6" # Minimum supported version
-- "3.8-dev" # Why not also test cutting edge?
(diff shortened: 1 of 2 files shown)


matthewkirby:
Since its an architecture bug resolution there really is not anything to review. So I am going to go ahead and merge it without a review.

PR #235, opened 2020-02-14
fixes #214 by removing, by default, empty bins from profile table

### LINKED ISSUE #214: Flag empty bins after polaraveraging
At the moment we are not checking if all the bins have galaxies in them and empty bins will cause the fit to crash. There are different ways to deal with it, but I'd suggest simply adding the number of galaxies in each bin in the profile table of the GC object. We can then decide the ignore the 'bad' bins for the fit at a later stage.

cristobal-sifon:
1) In `utils.compute_radial_averages`, also return number of objects per bin
2) In `polaraveraging.make_shear_profile`, added an `include_empty_bins` kwarg which is set to `False` by default, meaning the returned table only includes bins with at least 2 objects (bins with 1 object make the fit in `Example2_Fit...` crash, I guess because std=0)
3) Regardless of the value of `include_empty_bins`, the table returned by `polaraveraging.make_shear_profile` includes a new column `n_src` with the number of objects in each bin.

This way I did not need to introduce any changes to the example notebooks and we relieve the user from having to think about masking the table. If the user wants to do that manually (or perhaps combine tables, etc), this can still be done by setting `include_empty_bins` to `True`.


cristobal-sifon:
yes, I forgot to run pytest before requesting, sorry about that. On your suggestions, the first one was a mistake - I thought I checked I reverted to large numbers but hey. The second one, will do.

cristobal-sifon:
@combet all items checked!

PR #240, opened 2020-02-26
Issue/239/require astropy v4

### LINKED ISSUE #239: require astropy >=v4.0
...in `setup.py`, otherwise tests fail (see #236)

cristobal-sifon:
tries to fix #239. 

It will work if you don't have astropy installed at all. However, though it looks like it works, the version of astropy does not seem to be updated if you already have an older version of astropy installed.


cristobal-sifon:
I guess I created this branch from #235's branch. Will fix and PR again.

PR #241, opened 2020-02-26
require astropy v4.0

### LINKED ISSUE #239: require astropy >=v4.0

cristobal-sifon:
tries to fix #239.

It will work if you don't have astropy installed at all. However, though it looks like it works, the version of astropy does not seem to be updated if you already have an older version of astropy installed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,7 @@ setup(
         "Operating System :: OS Independent",
         "Programming Language :: Python"
         ],
-      install_requires=["astropy", "numpy", "scipy"],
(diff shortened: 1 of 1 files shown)


aimalz:
@cristobal-sifon Just to check, the problem happens with the [current master branch version of requirements.txt](https://github.com/LSSTDESC/CLMM/blob/master/requirements.txt), even though that already specifies the version?

cavestruz:
@cristobal-sifon is this ready for review?  In which case, we can add the tag PR: Ready for Review

Issue #205, opened 2019-11-25
Some ideas about demo_pipeline.ipynb


shenmingfu:
In `examples/demo_pipeline.ipynb` (`Example1_Fit_Halo_Mass_to_Shear_Catalog.ipynb`), there are several things that we might need to double check:

- [x] The x-axis unit in "Computing shear" is radian rather than degree.

- [x]  `e1` and `e2` in cluster catalog are considered as (reduced) shear rather than ellipticity. (We might need to adjust the definition of `e1` and `e2` in `mock_data`) 

- [x] In `cluster_toolkit`, the unit of mass is `$M_{sun} h^-1$`, but in CLMM we are using `M_{sun}`.

- [x] `polaraveraging.make_shear_profile` uses `compute_radial_averages` in `utils`, where the error bar is calculated by `std/n`. We might need to consider `std/numpy.sqrt(n)`.

- [x] We might need to make a note in the notebook: if the number of bins is too large, or the bin range is biased, it is possible that some bins have no data, and this could lead to `nan` (or some value divided by zero) in some statistical quantities.

combet:
Simply commenting on the items above: 
- Yes, the label in the shear profile should be radians and not degrees.
- `e1` and `e2` are indeed directly computed from the reduced shear in `mock_data.py`.  @shenmingfu, you're suggesting it's missing a factor 2 to convert to ellipticities? It does depend on the definition we take for ellipticities though, right? 
- Regarding the mass definition, I'm not sure I understand your point. CLMM's modeling uses `Msun/h` so the results of the fit will be in that unit, but that's all there is to keep in mind, no? I might be missing something...
- I agree the error bar on the bin average should be `std/sqrt(n)` and not `std/n` as it is now.
- The issue of empty bins in being addressed in issue #214.

shenmingfu:
Hi @combet , thank you. 

- Yes that's for the factor 2, and yes that depends on the definition of ellipticities. I find this [line](https://github.com/LSSTDESC/CLMM/blob/e7e0859658e813f01a639607cbcc4a5a37cc72a6/clmm/polaraveraging.py#L112) in `polaraveraging` so it is consistent.

[...]

PR #243, opened 2020-02-28
corrected errorbar: std/sqrt(n), not std/n

shenmingfu:
`polaraveraging.make_shear_profile` uses `compute_radial_averages` in `utils`, where the error bar is calculated by `std/n`. Corrected it to `std/numpy.sqrt(n)`. #205


shenmingfu:
@combet Thank you very much for pointing this out. I updated `polaraveraging.py` and `test_utils.py`.

combet:
Also asking @cavestruz for review as I do not have permission to merge into master.

Issue #233, opened 2020-02-13
Add noise on the cross shear in mock_data


combet:
At the moment, `mock_data.py` only adds noise on the tangential shear component if the user requests it, but leaves the cross shear strictly equal to zero. This needs to be updated so that both components are affected by noise. In `mock_data.py`
- [x] Draw `g_x` from a Gaussian distribution centered on zero
- [x] Adapt the expressions of `e1` and `e2`to account for `g_x`

This follows a [discussion](https://lsstc.slack.com/archives/C8D902XV3/p1569341896065200) on Slack a few month back.

combet:
Merged. Closing issue.

PR #246, opened 2020-03-13
Issue/233/mock crossshear noise

combet:
This addresses issue #233 and the original limitation of `mock_data.py` where the cross shear was strictly equal to zero by construction, even in the presence of shape noise.


examples/support/mock_data.py
```diff
@@ -147,15 +147,20 @@ def _generate_galaxy_catalog(cluster_m, cluster_z, cluster_c, cosmo, ngals, Delt
...
 
     # Compute ellipticities
     galaxy_catalog['posangle'] = np.arctan2(galaxy_catalog['y_mpc'], galaxy_catalog['x_mpc'])
-    galaxy_catalog['e1'] = -galaxy_catalog['gammat']*np.cos(2*galaxy_catalog['posangle'])
-    galaxy_catalog['e2'] = -galaxy_catalog['gammat']*np.sin(2*galaxy_catalog['posangle'])
+
+    galaxy_catalog['e1'] = -galaxy_catalog['gammat']*np.cos(2*galaxy_catalog['posangle']) \
+                           + galaxy_catalog['gammax']*np.sin(2*galaxy_catalog['posangle'])
+    galaxy_catalog['e2'] = -galaxy_catalog['gammat']*np.sin(2*galaxy_catalog['posangle']) \
+                           + galaxy_catalog['gammax']*np.cos(2*galaxy_catalog['posangle'])
```
shenmingfu: Hi @combet , sorry, could you check this line? I feel like there is a minus sign here (line 163)...

combet: @shenmingfu, you're right, thanks for checking this out. It's corrected now.

PR #256, opened 2020-03-31
Add magnification in modeling

marina-ricci:
I added a function to compute the magnification in clmm/modeling.py


combet:
Thanks @marina-ricci! I think it could be useful to:
- [ ] double check the magnification formula. The "Schneider book" has 4
Screenshot 2020-04-01 at 10 01 59
where I think |gamma| = sqrt(gamma_t^2 + gamma_x^2) (= gamma_t for a circularly-symmetric lens) is the shear, not the reduced shear as you have in the function. But I might have misunderstood.
- [ ] add a function to `test_modeling.py` to test the magnification function.

Also, I think it would be good to
- [ ] add the example use in the `demo_modeling_functionality` notebok

combet:
Hi @marina-ricci, just to let you know that PR #266 (fix issue when source is in front of the lens) that you were waiting for for this branch has now been merged in.

marina-ricci:
I started a new branch for this PR (issue_251_new) as I needed to re-start from scratch.

Issue #216, opened 2020-01-23
Make polaraveraging compatible with different data input


marina-ricci:
We would like to make polaraveraging more modular by being able to work with either ellipticities or shear or reduced shear. 
- We need to remove the warning and error messages based on the columns names.
- We need to add a  keyword for the user to specify the data type given in input e.g. 'ellipticities', 'shear' and 'convergence' or 'reduced  shear'.
- We need to operate the transformation to reduced shear if needed.

cavestruz:
Converting @marina-ricci 's comment to checkmarks

- [ ] We need to remove the warning and error messages based on the columns names.
- [ ] We need to add a keyword for the user to specify the data type given in input e.g. 'ellipticities', 'shear' and 'convergence' or 'reduced shear'.
- [ ] We need to operate the transformation to reduced shear if needed.

rherbonnet:
Thoughts for conversion function so that whatever the user inputs there will always be the same definition of the columns 'e1' and 'e2' in the astropy table (galcat) for the GalaxyCluster object

def helper_function(chi_1=None,chi_2=None,epsilon_1=None,epsilon_2=None, axis_ratio=None, position_angle=None, reduced_shear_1=None,reduced_shear_2=None,shear_1=None,shear_2=None, convergence=None):
          '''User can input whatever their catalogue gives and this will always return reduced shears 1 & 2, which can be put into an astropy Table for GalaxyCluster object'''
          return reduced_shears_1_and_2

combet:
Thanks for the helper function @rherbonnet. Could you please double check the implementation of the conversion between chi et epsilon?

I think there's a set of parentheses missing at the denominator of the conversion factor, i.e. `chi_conversion = 1./(1.+(1-(shape_1**2 + shape_2**2))**0.5)` from Bartelmann & Schneider (https://arxiv.org/pdf/astro-ph/9912508.pdf).
![Screenshot 2020-03-25 at 15 08 34](https://user-images.githubusercontent.com/20227809/77545302-85697c80-6eaa-11ea-9387-1fc6e60ffe89.png)

rherbonnet:
Yes that would be more readable, but since it's all additive between the parentheses, I don't think it matters. Right? I will change it to your suggestion, though

combet:
But there's the minus sign in front of the parenthesis I added, so no, it's not the same. Or am I missing something really obvious?

rherbonnet:
aaaah you're right, no I was missing something obvious :)
[...]

PR #258, opened 2020-04-01
Issue/216/polaraverage dc2

rherbonnet:
#216 New functionality is added so the user can define what ellipticity or shear (s)he is feeding the polaraveraging code. This should make it easier to switch between data and simulations input to the code.


clmm/polaraveraging.py
```diff
@@ -105,7 +105,8 @@ def compute_shear(cluster=None, ra_lens=None, dec_lens=None, ra_source_list=None
     if cluster is not None:
         required_cols = ['ra', 'dec', 'e1', 'e2']
         if not all([t_ in cluster.galcat.columns for t_ in required_cols]):
-            raise TypeError('GalaxyCluster\'s galaxy catalog missing required columns.')
+            raise TypeError('GalaxyCluster\'s galaxy catalog missing required columns.' +\
+                            'Do you mean to first convert column names?')
```
cavestruz, line 109: Helpful - thanks!

tests/test_utils.py
```diff
@@ -190,3 +190,46 @@ def test_convert_units():
...
                     truth, **TOLERANCE)
+
+    
+    
+def test_shape_conversion():
+    """ Test the helper function that convert user defined shapes into
+    epsilon ellipticities or reduced shear. Both can be used for the galcat in 
+    the GalaxyCluster object"""
+    
+    def ellipticities(q20,q11,q02):
```
cavestruz, line 201: So, I think this def ellipticities function should either live in utils.py or elsewhere because this could be generally useful.

tests/test_utils.py
```diff
@@ -190,3 +190,46 @@ def test_convert_units():
...
+        """Build ellipticties from second moments
+        """
```
cavestruz, line 203: Nitpicky, but can this also include:
`Parameters
    ----------
    q20 : array_like
        Second moment ????
    q11 : array_like
        ????
    q02: array_like
        ????
    Returns
    -------
    x1: array_like
        ??????
    x2: array_like
        ??????
    e1: array_like
        ??????
    e2: array_like
        ??????
`

tests/test_utils.py
cavestruz: also, can this be named, 
`def build_ellipticities(q20, q11, q02):
`

cavestruz:
For succinctness (and records), transferring my comments into a checklist:

- [ ] Modify function name from ellipticities: `def build_ellipticities(q20, q11, q02):`

- [ ]  Add Parameters and Returns to the docstrings of build_ellipticities

- [ ] Transfer build_ellipticities to utils.py (or elsewhere) because this could be generally useful.  But, for code coverage, we'll likely also want to add a very quick vtest_build_ellipticities() function in tests_utils.py

combet:
Thanks @cavestruz! Your points above have now been addressed.

PR #265, opened 2020-04-06
Issue/253/binning

rherbonnet:
Functionality has been added to allow the user to select equal occupation for bins, so roughly the same number of galaxies per radial bin. Also added a test in test_utils and an example in demo_polaraveraging_functionality

### DIFF
--- a/clmm/utils.py
+++ b/clmm/utils.py
@@ -117,7 +117,7 @@ def make_bins_with_same_ngal(separation_arr, separation_units, bin_units,\
     physical_bank = {"pc": u.pc, "kpc": u.kpc, "Mpc": u.Mpc}
     units_bank = {**angular_bank, **physical_bank}
 
-# Some error checking
+    # Some error checking
     if bin_units not in units_bank:
         raise ValueError(f"Input units ({bin_units}) not supported")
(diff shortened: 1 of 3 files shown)


combet:
The steps taken were:
- [ ] Add option `evenoccupation` in the `make_bins` function of `utils.py`
- [ ] Add corresponding test to `test_utils.py`
- [ ] Add usage example in `demo_polaraveraging_functionality.ipynb` (which was slightly reorganized to make things a bit clearer)

clmm/utils.py
```diff
@@ -62,6 +62,7 @@ def make_bins(rmin, rmax, nbins=10, method='evenwidth'):
         Binning method to use
         'evenwidth' - Default, evenly spaced bins between rmin and rmax
         'evenlog10width' - Logspaced bins with even width in log10 between rmin and rmax
+        'equaloccupation' - Bins with equal occupation numbers
```
cavestruz, line 65: Need to add new Parameter description:
`source_seps : array like
      radial distance of source separations`
(or some other wording for the docstrings)

combet: Thanks, done.

Issue #257, opened 2020-04-01
Error in modeling when z_source < z_lens


marina-ricci:
In clmm/modeling.py there is no warning or error when z_source < z_lens and the functions return non zero values. This is incorrect.

combet:
@marina-ricci, I looked into it and tried to fix it. Could you please checkout branch issue/257/... and also tentative PR #266 to see if this does what you were expecting?

combet:
Done in PR #266. Closing issue.

Issue #262, opened 2020-04-03
Adapt integration ranges in modeling.py to the user-defined r_proj argument 


combet:
Numerical integrations are performed by cluster toolkit to compute Sigma and DeltaSigma. The function being integrated, e.g rho or Sigma, needs to be defined on a domain that encompasses the `r_proj` array for which the user is requesting the evaluation. At the moment, this underlying domain has hard-coded values in `modeling.py`, and if it's not adapted to `r_proj`, this returns no clear error or warning. This should be changed to automatically adapt to whatever radial range is covered in r_proj.
- [ ] in `modeling.py` change any hard-coded instance such has `sigma_r_proj = np.logspace(-3, 4, 1000)` (line 271) to something like `sigma_r_proj = np.logspace(np.log10(r_proj[0])-1 , np.log10(r_proj[-1])+1, 1000)`
- [ ] If `r_proj` requests too small values, say < 1e-5, return a warning?

(Extending the range by one order of magnitude, like in the example above, might be overkilling it?)

combet:
Adding to the above, looking at where cluster-toolkit complained, the smallest radial range for the integrand needs to be at least that of the user-defined range when calling `predict_excess_density`, etc. While working OK with this minimal radial range, the validation tests, comparing to NumCosmo, results were failing for rtol=1.e-8. While passing with rtol=1e-7, I opted for keeping the current 1e-8 requirement and adjusting the integration range for it to pass. That led to:
- [ ] `sigma_r_proj = np.logspace(np.log10(r_proj[0])-1, np.log10(r_proj[-1])+1, len(r_proj)*10)`  in `predict_excess_surface_density`;
- [ ]  raising an error for very small radii (<1.e-11 Mpc/h) in DeltaSigma, where the kernel was consistently crashing.

combet:
This has been merged, so closing for now. Might be reopened if required.

combet:
@rherbonnet just noticed that the modification cannot handle the case when r_proj is a float and not array-like. A small fix needs to be implemented in `modeling.py`. Reopening this issue until fixed.

rherbonnet:
Just pushed the fix. Instead of using if statements, I used np.array(r_proj).size assuming and hoping that the user is not going to give a more than 1D array of projected distances

combet:
[...]

PR #266, opened 2020-04-07
Set shear and convergence to zero if source is in front of the lens

combet:
This fixes the behavior noted by @marina-ricci in #257 where `modeling.py` returns negative values for shear and convergence when the source is in front of the cluster. These values should be set to zero.
- [x] In `modeling.py`, code was added to check z_source w.r.t z_cluster and returns zero accordingly.
- [x] A warning is also issued to tell the user this happened
- [x] A new test was added to `test_modeling.py` to capture this new feature.


rherbonnet:
Just a thought that here also the critical surface density should actually be set to infinity because it's
Sigma_crit = (c^2/4pi G) (1/D[ol]) (1/beta_s)
where
beta_s = max(0, D[ls]/D[os])
and D the angular diameter distance  between o(bserver), l(ens) and s(ource).

combet:
Thanks @rherbonnet! You're absolutely right and I should have started with `Sigmacrit`. So I've now enforced `Sigmacrit = np.inf` if the source is in front of the cluster. From that, convergence=0 and shear=0 follow naturally, so I removed the code I added in those functions as it became redundant. Also added a test for sigmacrit in `test_modeling.py`.

rherbonnet:
I've added extra lines to ensure kappa and gamma are zero, for the few pathological cases @combet noticed where dividing by np.inf did not return 0 (1e300/np.inf=0, 1.e400/np.inf =nan).
Now it's three ugly if statements, so if anyone knows a good way of performing an operation on an object that's either a float or an array, I would be keen to know.

marina-ricci:
@rherbonnet I faced the same issue =) You could try ` np.nan_to_num`. It works with scalar and array and return 0 or +- inf, according to the operation. I checked that  `np.nan_to_num( 1.e400 / np.inf) ` returns 0.

marina-ricci:
@combet @rherbonnet You can have a look at  the version of `modeling.py` I have here: https://github.com/LSSTDESC/CLMM/pull/256/files. It works well with only few lines added.

rherbonnet:
Thanks @marina-ricci, I couldn't break your solution, so I implemented it. The only problem I can now think of is the weird case of a user giving a redshift array with an infinity or nan in it. Then np.nan_to_num automatically makes those entries 0 without raising a warning. Would we want that? 
There is now a print statement mentioning that tells the user when there are z_sources<z_cluster. I was thinking of adding a similar thing for nonfinite z_source values.
It's probably not going to happen, but I was thinking of cases that might break the code.
What do you think?

combet:
Thanks @marina-ricci, your solution is really neat and I couldn't break it either.

@rherbonnet, I think a warning for non finite zsource values would be appropriate, yes. Hopefully it won't be needed, but we've seen weirder things.

rherbonnet:
I tried some more weird stuff and you actually get a lot more _RuntimeWarning: invalid value encountered_ because z_source is used all over modeling.py
It might be a bit much to add an if statement everywhere. The user already get the runtime warnings

combet:
Right, I hadn't realised there was RuntimeWarning... It might be enough, then? I agree that crowding the code with if statements for a situation that should very rarely arise (and that gets a warning anyway) may not be the right approach. Leave it like this then?

rherbonnet:
I just pulled before committing and I hit an error. @combet I found a case in which your solution of #262 crashes. 
You added:
_sigma_r_proj = np.logspace(np.log10(np.min(r_proj))-1, np.log10(np.max(r_proj))+1, len(r_proj)*10)_
But now I get the error
_TypeError: object of type 'float' has no len()_
when running test_modeling.py, because one of the tests has r_proj = 1 and "1" has no length

combet:
Thanks for catching that! I overlooked that r_proj could simply be a float and not array-like. I just reopened issue #262 to highlight that this needs fixing in `modeling.py`. Branch issue/262 still exists so if you'd like to try to build a solution, it might be better to do it there. I'll look into it on Monday in any case.

clmm/modeling.py
```diff
@@ -484,7 +495,8 @@ def predict_convergence(r_proj, mdelta, cdelta, z_cluster, z_source, cosmo, delt
 
     if z_src_model == 'single_plane':
         sigma_c = get_critical_surface_density(cosmo, z_cluster, z_source)
-        kappa = sigma / sigma_c
+        kappa = np.nan_to_num(sigma / sigma_c)
```
cavestruz, line 498: Can we explicitly add the argument for nan (also for readability):
`kappa = np.nan_to_num(sigma / sigma_c, nan=0.0)`

clmm/modeling.py
```diff
@@ -542,7 +558,8 @@ def predict_reduced_tangential_shear(r_proj, mdelta, cdelta, z_cluster, z_source
                                     z_src_model)
         gamma_t = predict_tangential_shear(r_proj, mdelta, cdelta, z_cluster, z_source, cosmo,
                                            delta_mdef, halo_profile_model, z_src_model)
-        red_tangential_shear = gamma_t / (1 - kappa)
+        red_tangential_shear = np.nan_to_num(np.divide(gamma_t , (1 - kappa)))
```
cavestruz, line 561: (1) Does the gamma_t/(1-kappa) no longer work here?
Can we explicitly add nan kwarg for readability:
`red_tangential_shear = np.nan_to_num(gamma_t / (1 - kappa), nan=0.0)`

clmm/modeling.py
```diff
@@ -417,10 +424,10 @@ def predict_tangential_shear(r_proj, mdelta, cdelta, z_cluster, z_source, cosmo,
     delta_sigma = predict_excess_surface_density(r_proj, mdelta, cdelta, z_cluster, cosmo,
                                                  delta_mdef=delta_mdef,
                                                  halo_profile_model=halo_profile_model)
-
     if z_src_model == 'single_plane':
         sigma_c = get_critical_surface_density(cosmo, z_cluster, z_source)
-        gammat = delta_sigma / sigma_c
+        gammat = np.nan_to_num(delta_sigma / sigma_c)
```
cavestruz, line 429: Explicitly add:
`gammat = np.nan_to_num(delta_sigma / sigma_c, nan=0.0)`

cavestruz: Quick question - I think sigma_c can be 0, so gammat can also be infinity.  In this case, the default for nan_to_num is to make that some "large number".  Should we explicitly assign what nan_to_num does with infinity?

clmm/modeling.py
```diff
@@ -360,7 +362,12 @@ def get_critical_surface_density(cosmo, z_cluster, z_source):
     d_s = angular_diameter_dist_a1a2(cosmo, aexp_src, 1.0)
     d_ls = angular_diameter_dist_a1a2(cosmo, aexp_src, aexp_cluster)
 
-    sigmacrit = d_s / (d_l * d_ls) * clight_pc_s * clight_pc_s / (4.0 * np.pi * gnewt_pc3_msun_s2)
+    beta_s = np.maximum(0, d_ls/d_s)
+    sigmacrit = clight_pc_s * clight_pc_s / (4.0 * np.pi * gnewt_pc3_msun_s2) * 1/d_l * np.divide(1., beta_s)
```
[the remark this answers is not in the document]
combet: Done.

clmm/modeling.py
```diff
@@ -360,7 +362,12 @@ def get_critical_surface_density(cosmo, z_cluster, z_source):
...
+    
+    mask = (np.array(z_source)<=z_cluster)
+    if np.sum(mask)>0:
+        warnings.warn(f'Some source redshifts are lower than the cluster redshift. Returning Sigma_crit = np.inf for those galaxies.')
```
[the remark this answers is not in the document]
combet: I think keeping "Sigma_crit" for the specific case of the on-screen warning is more explicit than sigma_c (which is now consistently used throughout modeling.py).

clmm/modeling.py
```diff
@@ -429,6 +436,10 @@ def predict_tangential_shear(r_proj, mdelta, cdelta, z_cluster, z_source, cosmo,
     #                               'distribution of redshifts in each radial bin')
     else:
         raise ValueError("Unsupported z_src_model")
+    
+    if np.sum(np.array(z_source)<=z_cluster):
```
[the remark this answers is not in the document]
combet: Yes and done.

combet:
Thanks @cavestruz for the review and suggestions. I addressed the three points. But then, looking at your last comment and looking at the code again, I actually have mixed feelings about `nan_to_num`:
- If the source is in front of the lens, then `sigma_c =np.inf` and in the case where Sigma and DeltaSigma are finite, then `kappa = gamma_t = gt = 0` follow naturally.
- Now, if for some reason, kappa, gamma_t and gt are `nan` or `np.inf`, I think we may want to keep that information rather than hide it behind a 'normal' numerical value as it means that something went really wrong and we want to be aware of when that happens.

What do you think? Also pinging @marina-ricci and @rherbonnet.

combet:
Also, having `nan=0.0` as keyword of `np.nan_to_num` appears to make Travis unhappy. Tests are passing on my machine though. Probably linked to the numpy version as it requires `numpy > 1.17` for this option to work... Not sure how to fix that for Travis.

marina-ricci:
> Thanks @cavestruz for the review and suggestions. I addressed the first three points. But then, looking at your last comment and looking at the code again, I actually have mixed feelings about `nan_to_num`:
> 
>     * If the source is in front of the lens, then `sigma_c =np.inf` and in the case where Sigma and DeltaSigma are finite, then `kappa = gamma_t = gt = 0` follow naturally.
> 
>     * Now, if for some reason, kappa, gamma_t and gt are `nan` or `np.inf`, I think we may want to keep that information rather than hide it behind a 'normal' numerical value as it means that something went really wrong and we want to be aware of when that happens.
> 
> 
> What do you think? Also pinging @marina-ricci and @rherbonnet.

I think the good way to use np.nan_to_num is to do : 
np.nan_to_num(x, nan=np.nan, posinf=np.inf, neginf=-np.inf), so when x = 1/np.divide(1., 0), it return 0 but when x = nan or inf, it return nan or inf.

cavestruz:
That sounds reasonable to me.

combet:
Thanks @marina-ricci, but now I feel like I'm missing something obvious and sorry if that's the case. In your example, I don't see the point of using `nan_to_num` given that `nan` and `inf` will remain `nan` and `inf`. x = 1/np.divide(1., 0)=0.0 and does not need a `nan_to_num` to remain so. So what does `nan_to_num` brings?

combet:
So, I've opted to use `nan_to_num` using @marina-ricci suggested syntax. I'm still not entirely sure it's needed when used this way but it does not do anything wrong and might bring some extra safeguard anyways. Tests are all passing, master has been merged. Re-requesting review.

Issue #260, opened 2020-04-02
Check and solve commented tests in test_polaraveraging


combet:
There are a bunch of commented out tests in the `test_polaraveraging.py` saying that these rais a warning when running pytest. This should be understood and/or corrected/cleaned out accordingly. (First mentioned in #164).

pennalima:
## utils.py

- Function compute_radial_averages: the standard deviation of the mean was ill-defined, it lead to a division by zero for empty bins. Solution implemented in this branch: each empty bin has its weight set to one.

## test_polaraveraging.py

1) Removed the tests named "Some additional edge cases" because they were failing  since the variable TOLERANCE was missing, and just below there are the same edges tests.
2) Removed the test of the function compute_lensing_angles_flatsky when RA_s = DEC_s = 0.0.The angular separation in this example is > 1. It is not necessary to use this function here. There is already another test checking if the warning for ang. separation > 1 is thrown. 
3) Removed the test for RA_s = RA_l, DEC_s = DEC_l. As it was commented, this is not an error.
4) Test: "Failure when lens and a source share an RA and a DEC". This was not working because the function arctan2(y,x) does not necessarily returns zero when x=y=0. From the manual:
  - If y is +0 (-0) and x is -0, +pi (-pi) is returned.
  - If y is +0 (-0) and x is +0, +0 (-0) is returned. 
To work around this phi is set to zero when angsep == 0. This modification was done in _compute_lensing_angles_flatsky. 
5) When the angular separation RA_s - RA_l is n*2*pi + epsilon, the function  _compute_lensing_angles_flatsky was resulting in a large deltax, this was due the
fact that it was computed  (RA_s - RA_l)*cos(dec_l) and then setting it to (-pi,pi).
Now it computes (RA_s - RA_l) and sets it to be in (-pi,pi) and then multiply
the result by cos(dec_l).
6) Updated the tests checking the case RA_s - RA_l is n*2*pi + epsilon and removed unnecessary ones.

Please check if all these points make sense to you.

combet:
This has been merged - closing issue.

combet:
This has been merged - closing issue

PR #269, opened 2020-04-17
Removed old tests, fixed some and minor improvements

vitenti:
Removed old tests, fixed some and minor improvements in compute_radial_averages and _compute_lensing_angles_flatsky.

compute_radial_averages

- std_sqrt_n was emmiting a warning due a division
  by zero. This happens for empty bins. Added
  a simple statement that put the weight to one
  and avoids division by zero.

_compute_lensing_angles_flatsky

- When source and lens have equal angles the atan2
  function returns phi = +/-pi due its evaluation
  of signed zeros. Added a statement that sets
  phi = 0 when sepangle == 0 (phi is ill defined
  when sepangle -> 0).

- Removed tests that emmited warnings due to large
  angle separations. Added support for source and lens
  angles separated by n*2*pi+epsilon and included
  tests to check it.

See more details in issue #260 .


shenmingfu:
One thing I realized is that, in `Python 3.7.4`, if the one bin is empty, the `mean` from `binned_statistic` would be nan. For example:
>\>\>\> binned_statistic([0,10,100],[0,10,100],statistic='mean',bins=[2,5,90])
BinnedStatisticResult(statistic=array([nan, 10.]), bin_edges=array([ 2., 5., 90.]), binnumber=array([0, 2, 3]))

However, for `std`

>\>\>\> binned_statistic([0,10,100],[0,10,100],statistic='std',bins=[2,5,90])
BinnedStatisticResult(statistic=array([0., 0.]), bin_edges=array([ 2., 5., 90.]), binnumber=array([0, 2, 3]))

Therefore, we might need to consider this in the mean value.

vitenti:
> `binned_statistic`

That is the behavior for 'mean' described in the documentation. However, for 'std' they do not specify the behavior in the case of empty bins. It may be a good idea to get the indices of NaN in mean and use them to replace the NaN by 0 in 'mean' and to assert the same indices are 0 also in 'std'. Maybe there is a better solution, how exactly these bins are going to be used later?

shenmingfu:
> > `binned_statistic`
> 
> That is the behavior for 'mean' described in the documentation. However, for 'std' they do not specify the behavior in the case of empty bins. It may be a good idea to get the indices of NaN in mean and use them to replace the NaN by 0 in 'mean' and to assert the same indices are 0 also in 'std'. Maybe there is a better solution, how exactly these bins are going to be used later?

This is a good point. I think the bins are mainly for the distance to the center of the cluster. Maybe we can try use 'count' in binned_statistic first, then select those bins with output 0 by indices, locate them by indices and replace those results (NaN) in 'mean' and 'std'.

tests/test_polaraveraging.py
```diff
@@ -90,21 +85,15 @@ def test_compute_lensing_angles_flatsky():
     ra_l, dec_l = 161.32, 51.49
     ra_s, dec_s = np.array([161.29, 161.34]), np.array([51.45, 51.55])
     thetas, phis = pa._compute_lensing_angles_flatsky(ra_l, dec_l, ra_s, dec_s)
+
     testing.assert_allclose(thetas, np.array([0.00077050407583119666, 0.00106951489719733675]),
```
shenmingfu, line 89: I'm a bit confused here. Can I ask where these numbers came from? Are they from some special cases? Thanks.

vitenti: I didn't write those tests, but I imagine that they are just sanity tests.

clmm/utils.py
```diff
@@ -40,7 +40,9 @@ def compute_radial_averages(xvals, yvals, xbins, error_model='std/sqrt_n'):
         yerr = binned_statistic(xvals, yvals, statistic='std', bins=xbins)[0]
     elif error_model == 'std/sqrt_n':
         yerr = binned_statistic(xvals, yvals, statistic='std', bins=xbins)[0]
```
shenmingfu, line 42: Maybe we can try this here to avoid that `mean` in `binned_statistic` gives NaN when some bin is empty.
```suggestion
        count = binned_statistic(xvals, yvals, statistic='count', bins=xbins)[0]
        count_zero = count==0
        
```

cavestruz: @vitenti - I'd be happy to make some of @shenmingfu  's suggested changes in a co-work if they sound reasonable to you.  Let me know.

vitenti: Yes, it seems reasonable to standardize the output with zero instead of NaN for empty bins components. I just modified this branch accord to @shenmingfu suggestions.

PR #272, opened 2020-04-30
Issue/263/massfit with zdistrib

payerne:
A new notebook `Example3...` added to the examples section demonstrates how to use CLMM for halo mass fitting when background galaxies are distributed in redshift.


marina-ricci:
I tested the notebook and it worked well for me!  I have the following minor comments : 
- Could you consider not including shape noise in the noisy model to only see the effect of the noise on redshifts? 
- separate the import `from sampler import fitters` to highlight the fact it's an CLMM module
- In the mass fitting part, explain why you fit for logM and not M directly and how you compute the error on the mass. 
- When you print the results of the mass fit, it would be good to recall the input mass. 
-They are some typos in the text (e.g. ` fornoisy`, ` true for noisy data for noisy data `...) 
- Better explains what is done in the last part (also the titles "shear to mass" are not very informative).
- It would be nice to add a conclusion after the last plots as I think they are the most interesting.
- Overall the clarity of the code can be improved as it is a bit hard to read through.

payerne:
Thanks @marina-ricci for the review, I adressed all your points and hope the notebook is clearer now.

PR #276, opened 2020-05-08
Make GCData the main data object to be used by the code

m-aguena:
Right now we just need make GCData object have all the properties of an astropy table and replace all instances of the use of astropy tables in the code by GCData.


m-aguena:
@combet `meta` is a property that astropy Tables have and  they are somewhat restrictive on the values you can pass to it. I created `metadata` so we could be a little more flexible with it and also because I thought it was a better self explained name (a bit of a nitpick on my part). However, right now I am not sure it `metadata` is actually expanding the possibilities of values it can assume, I have not really tested. So maybe, I can just remove it for simplicity sake. What do you think?

combet:
@m-aguena, thanks for the clarification. I do not have strong feelings either way, but maybe a slight preference for removing any redundancy, and therefore keeping `meta` only. I'll leave the final decision to you. With the merge that happened last week, there are now 2 small conflicts with `master` to fix. Please, re-request a review at your convenience.

PR #278, opened 2020-05-13
Add magnification in modeling

marina-ricci:
This PR aims at including the modeling of the magnification in CLMM : 

- [x] I added a function predict_magnification in `modeling.py`

- [x] I added an example in `demo_modeling_functionability.ipynb`

- [x] I added tests in `test_modeling.py` also a based on the new `radial_profiles.txt` file from numcosmo, that now includes a column with magnification.


marina-ricci:
The tests are failing because there are some differences between the old and new `radial_profiles.txt`. I need some feeback on that before the PR is ready.

PR #286, opened 2020-06-23
Issue/125/deltasigma from data

combet:
This addresses the question of computing a DeltaSigma profile from the data. The following has been done:
- New option was added to `compute_tangential_and_cross_components`. If `is_deltasigma=True` then results are multiplied by `Sigma_c` before being returned. This also means that `modeling` needs to be imported.
- A few more lines in `test_polaraveraging.py`, making sure errors are raised properly when redshifts or cosmology are missing when using the new option.
- New section (4.3) added to the `demo_polaraveraging_functionality` notebook to exemplify the usage of this option.


combet:
Thanks @m-aguena. I modified the documentation of the function to clarify that this is only an ESTIMATE of `DeltaSigma` when working with ellipticities or reduces shear. It is only equivalent to `DeltaSigma` if the user is directly dealing with the shear, which can happen when working from simulations. 

There are references where we do see the explicit multiplication of, say, tangential ellipticities and Sigma_c to provide the estimate of DeltaSigma, but this is generally done in stacking analyses, where weights and sum over all lens-src pairs are also involved, which I think make the matter more confusing. 

Let me know if the improved description in the documentation is sufficient.

Issue #282, opened 2020-05-29
Modeling structure


vitenti:
## CLMM Modeling structure

Currently, we have a functional structure where all functions receive the full description of what they need to compute. This is partially done through the cosmo object (that can be either an astropy LCDM model or a dictionary) that contains all information about the background cosmology. The parameters related to the halo density profile are all passed one by one in each function. Below, we have the current structure: the  miscellaneous functions do not depend on cosmo nor halo density parameters,  and the other functions depend on the halo density parameters, background cosmology "cosmo" or both.

#### Miscellaneous utilities:

- def cclify_astropy_cosmo(cosmoin):

- def astropyify_ccl_cosmo(cosmoin):

- def _get_a_from_z(redshift):

- def _get_z_from_a(scale_factor):

- def get_reduced_shear_from_convergence(shear, convergence):

#### Halo density profile:

- def get_3d_density(r3d, mdelta, cdelta, z_cl, cosmo, delta_mdef=200, halo_profile_model='nfw'):

- def predict_surface_density(r_proj, mdelta, cdelta, z_cl, cosmo, delta_mdef=200, ...

- def predict_excess_surface_density(r_proj, mdelta, cdelta, z_cl, cosmo, delta_mdef=200, ...

#### Background cosmology:

- def angular_diameter_dist_a1a2(cosmo, a1, a2=1.):

- def get_critical_surface_density(cosmo, z_cluster, z_source):

#### Background cosmology and Halo density profile:

- def predict_tangential_shear(r_proj, mdelta, cdelta, z_cluster, z_source, cosmo, delta_mdef=200, ...

- def predict_convergence(r_proj, mdelta, cdelta, z_cluster, z_source, cosmo, delta_mdef=200, ...

- def predict_reduced_tangential_shear(r_proj, mdelta, cdelta, z_cluster, z_source, cosmo, ...

- def predict_magnification(r_proj, mdelta, cdelta, z_cluster, z_source, cosmo, delta_mdef=200, ...

[...]

PR #291, opened 2020-07-10
Issue/285/reorganization

aimalz:
I made a first stab at reorganizing where CLMM's cosmology-dependent and cosmology-independent functions live.  The initial plan was that there would be four modules, for all possible combinations of cosmology- and data- dependence.  Cosmology dependence is easy to identify -- does the function take a `cosmo` object? -- but it seems that data-dependence is really defined by whether the function needs a `GCData` object.  I reorganized where the functions live based on these definitions:
- `dataops.py` (formerly `polaraveraging.py`) for `GCData`-dependent functions
- `hybrid.py` for functions of `cosmo` and `GCData` objects
- `modeling.py` for `cosmo`-dependent functions
- `utils.py` for functions that don't touch either type of object

Note that `dataops.py` now contains only one function, and almost everything from `polaraveraging.py` ended up in `utils.py` (which I'd sort of like to rename).

In particular, I think this sets us up well for implementing @vitenti 's suggestion (#282) to build an object on top of these functions, but for both the `cosmo`-dependent and the `GCData`-dependent functions.

**This is a draft PR because the branch is intended to be a starting point for discussion!**  Please share your thoughts in the comments.

### DIFF
--- a/clmm/__init__.py
+++ b/clmm/__init__.py
@@ -1,10 +1,13 @@
 """ CLMM is a cluster mass modeling code. """
 from .gcdata import GCData
 from .galaxycluster import GalaxyCluster
+
+from .dataops import compute_tangential_and_cross_components
 from .hybrid import make_binned_profile, convert_units
-from .polaraveraging import compute_tangential_and_cross_components
-from .utils import compute_radial_averages, make_bins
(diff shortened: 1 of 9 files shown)


combet:
This had been addressed by the defintion of the cosmological object (#311) and by PR #321. Can be closed.

Issue #284, opened 2020-06-12
mock_data module: double check everything, improve/update options and improve documentation


combet:
Need to double check everything in `mock_data` and improve documentation. Especially

- [x] Check/improve documentation, in particular clarify comoving versus proper distance

- [ ] Improve shape noise? Is adding a random gaussian noise to `gt` equivalent to drawing intrinsic `e1` and `e2` and add those to the shear using the `compute_lensed_ellipticities` function we now have available in `utils.py`  ? (Now issue #304 )

- [x] Double check and fix 'ztrue' versus 'zobs' potential bug

- [ ] Write test functions for mock_data (Now issue #303 )

@marina-ricci, please add anything else you notice

- [x] Give the option to pass a density of source galaxies, or a number of galaxies and a box size

- [x] Give the box size as an argument and not a hard coded value.

- [x] Give the type of profile as an argument.

- [ ] Give the source redshift distribution as an argument

combet:
Thanks @marina-ricci for the list of things to add/improve. 

Just a  clarification regarding your last point (source redshift distribution): at the moment, the user uses either e.g. `zsrc=0.8` for a single source plane or `zsrc='chang13'` to get a sources distributed in redshift. What are you suggesting more specifically?

marina-ricci:
Please check that this issue is addressed : 
 - The redshift cut in mock_data should be done differently to avoid error messages because of z_source<z_cluster.

combet:
Did it this morning - should be fine now. Either the user provide a zsrc_min or it defaults to z_cluster + 0.1

combet:
@marina-ricci , @pennalima, I've opened a PR for this issue as all has been adressed but for the tests. This is a bigger task so I've opened a new issue (#303) that focusses solely on writing tests for `mock_data` as this is not as straightforward as the rest.

combet:
>     * [ ]  Give the source redshift distribution as an argument

@marina-ricci, I think you meant passing a user-defined distribution as argument and we concluded at some point that this should be a enhancement for later, right ?

Issue #268, opened 2020-04-15
list of minor things that need to be modified


marina-ricci:
This is a list of minor things that need to be modified. Once the list is long enough we can consider making a PR. Some of these might already been fixed by previous PR.

- [x] The documentation of function `compute_radial_averages` in `utils.py` was not updated after PR# 255

- [x] The redshift cut in `mock_data` should be done differently to avoid error messages because of z_source<z_cluster. 
- [x] The boxsize in `mock_data` is set to be 4 Mpc/h * 4 Mpc/h, instead what is given in the doc. This is starnegely defined and should be more flexible (e.g. giving source density instead of number of galaxies) 
- [x] `compute_shear()` returns a catalog with bad column names (ra, e1, e1 instead of theta, gt, gx)
- [x] `compute_shear((cluster_ra, cluster_dec, gal_cat['ra'], gal_cat['dec'], gal_cat['e1'], gal_cat['e2']))` does not work
- [x] `cl.plot_profiles` => the documentation should mention that is returns both 'fig' and 'ax'
- [x]  `sampler.py` is quite cryptic. I do not see the need to make an if statement in `def scicurve_fit`
- [x]   `cl.plot_profiles` is not plotting the cross-shear component.

combet:
Updates in the code since this issue was open:
- `cl.plot_profiles` not plotting the cross-shear component has been fixed as a side product in #248 
- `compute_shear` has been renamed `compute_tangential_and_cross_components` and modified for more flexibility in #248.
- Anything about `mock_data` should be addressed in #284

marina-ricci:
- [ ] The documentation of function `compute_radial_averages` in `utils.py` was not updated after PR# 255
- [ ] `cl.plot_profiles` => the documentation should mention that is returns both 'fig' and 'ax'
- [ ]  `cl.plot_profiles` =>not working when there is no "g_t" column (e.g. becaus euser changed the name to `g_tan`)
- [ ] In the `demo_polaraveraging_functionality` notebook documentation : in Section " 4.2. User-defined naming scheme " change "shear_profile" to "reduced_shear_profile"

PR #297, opened 2020-07-24
Fixing small bugs listed in issue #268

marina-ricci:
This PR addresses some of issue #268. 

- [x] The documentation of function `compute_radial_averages` in `utils.py` was not updated after PR# 255
- [x] `cl.plot_profiles` => the documentation should mention that is returns both 'fig' and 'ax'
- [x]  `cl.plot_profiles` =>not working when there is no "g_t" column (e.g. because user changed the name to `g_tan`)
- [x] In the `demo_polaraveraging_functionality` notebook documentation : in Section " 4.2. User-defined naming scheme " change "shear_profile" to "reduced_shear_profile"
- [x] compute_tangential_and_cross_components() returns a catalog with bad column names (ra, e1, e1 instead of theta, gt, gx)
- [x] compute_tangential_and_cross_components(cluster_ra, cluster_dec, cl.galcat['ra'], cl.galcat['dec'], cl.galcat['e1'], cl.galcat['e2']) 

#268 also raised issues with mock data. These specific points have been moved to issue #284


combet:
>     * compute_tangential_and_cross_components(cluster_ra, cluster_dec, cl.galcat['ra'], cl.galcat['dec'], cl.galcat['e1'], cl.galcat['e2'])

For the last point above, the argument are actually keywords, so this approach (not relying on a GC object) does work but with the following syntax:
`compute_tangential_and_cross_components(ra_lens=cluster_ra, dec_lens=cluster_dec, 
                                                          ra_source_list=cl.galcat['ra'], dec_source_list=cl.galcat['dec'], 
                                                          shear1=cl.galcat['e1'], shear2=cl.galcat['e2'])`

combet:
>     * compute_tangential_and_cross_components() returns a catalog with bad column names (ra, e1, e1 instead of theta, gt, gx)

Fixed by casting the `shear1`, `shear2`, `ra_source_list` and `dec_source_list` array-like quantities to numpy arrays. In the example of the last bullet point, these were astropy columns, with some names, and these names were propagated to the output in a manner I do not understand. Anyhow, forcing those to be numpy arrays fixes this issue.

combet:
> * `cl.plot_profiles` =>not working when there is no "g_t" column (e.g. because user changed the name to `g_tan`)

If I remember, we agreed that we should spend too much effort in making `plot_profiles` a very generic function given that this is aimed simply at quick checks. I'd suggest simply returning an error explaining that columns named 'g_t' and 'g_x' are required for it work.

shenmingfu:
I have added some description in `plot_profile()`: it returns both 'fig' and 'axes'.
For the `gt` and `gx` columns in the astropy Table `cluster_profile`, I used two different ways to check: comparing if the strings are equal, or if the column names exist. This depends on how we use the function, and one of them can be commented out.
Before "git push origin branchname" I did "git pull origin master". (https://github.com/LSSTDESC/CLMM/blob/issue/300/numcosmo_ccl_installdocs/CONTRIBUTING.md)
Did I do it correctly?
Please let me know if there's anything I need to adjust. Thanks.

combet:
Thanks @shenmingfu, this looks good. I'm changing the label to "Ready for review".

clmm/utils.py
```diff
@@ -30,6 +30,10 @@ def compute_radial_averages(xvals, yvals, xbins, error_model='std/sqrt_n'):
         Error on the mean y value in each bin. Specified by error_model
     n : array_like
         Number of objects in each bin
+    binnumber: 1-D ndarray of ints
```
cavestruz, line 33: Would it be more self-consistent to keep this as xbins being returned?  (there is redundancy between xbins or binnumber, yes?)

clmm/polaraveraging.py
```diff
@@ -159,14 +162,14 @@ def compute_tangential_and_cross_components(cluster=None,
```
aimalz, line 162: Suggestion re: ensuring all inputs are converted to arrays, since we probably have to do this a lot with astropy tables, maybe we should make a function in `utils.py` to do this check for all inputs in the first line of `_compute_*_shear()`, something like this:
```
def force_arrs(*args):  
    outargs = []  
    for arg in args:  
        outargs.append(np.array(arg))  
    return(outargs)
```

clmm/polaraveraging.py
```diff
@@ -159,14 +162,14 @@ def compute_tangential_and_cross_components(cluster=None,
 
     # Compute the lensing angles
     if geometry == 'flat':
-        angsep, phi = _compute_lensing_angles_flatsky(ra_lens, dec_lens, ra_source_list,
-                                                      dec_source_list)
+        angsep, phi = _compute_lensing_angles_flatsky(ra_lens, dec_lens, np.array(ra_source_list),
```
cavestruz, line 165: The docs mention ra_source_list as "array like".  Minor comment/recommendation:  recommend - either specify numpy array as type, thus avoiding the need to wrap as a numpy array, and/or rename "ra{,dec}_source_list" -> "ra{,dec}_sources" because the "_list" implies list type.

clmm/utils.py
```diff
@@ -30,6 +30,10 @@ def compute_radial_averages(xvals, yvals, xbins, error_model='std/sqrt_n'):
         Error on the mean y value in each bin. Specified by error_model
     n : array_like
```
cavestruz, line 31: Minor suggestion: rename to n -> num_objects

aimalz (approved):

clmm/utils.py
combet: Unless I misunderstood your point, no, they provide different information. `xbins`gives the bin edges. `binnumber` gives, for each galaxy, the indices of the bin it falls into. This is needed is the user requires to keep track of which galaxy belong to which bin.

clmm/utils.py
combet: Done

clmm/polaraveraging.py
combet: I removed the misleading `_list` from the variable names and explicitly mention that these should be arrays and not array-like.

clmm/polaraveraging.py
combet: Right, this is another way to approach this but given the small cope of this particular issue, I went with Camille's suggestion, removed the misleading `_list` from the variable name and updated the documentation stating that these should be arrays, not array-like. Letting the user take care of this, rather than making things more complex in the code.

PR #298, opened 2020-07-24
update numCosmo notebook to be compatible with numCosmo 0.15.1

boutigny:
The fix to make it compatible with numCosmo 0.15.1 is pretty straightforward.
Here are the modified lines (maybe easier to see than with the diff tool:

dp              = Nc.HaloDensityProfileNFW.new (Nc.HaloDensityProfileMassDef.MEAN, mass_Delta)
dp.props.MDelta = cluster_mass / cor_factor / cosmo.h ()
dp.props.cDelta      = cluster_concentration


combet:
- I've updated one of the comment to specify that CLMM in that example is using `cluster-toolkit` as the modeling backend (default).
- Also cleared out all the notebook outputs - these are often a source of conflict.
- Finally, this was a useful NB to have but I don't know if it should remain in the main example directory. TBD...

aimalz:
This is included in #299.

Issue #287, opened 2020-06-26
Re-document function in cluster_toolkit_patches


m-aguena:
The function `cluster_toolkit_patches._patch_zevolution_cluster_toolkit_rho_m` is making the right adjustments to use `cluster_toolkit`, but its name and documentation is based on old assumptions and is causing confusion. It must be updated.

m-aguena:
Has been solved, but changes will be implemented via #282.

Issue #277, opened 2020-05-08
Modeling: cross-check CCL, NumCosmo, Colossus...


pennalima:
The (probably) new structure concerning modeling.py must guarantee that all possible codes are cross-checked.

aimalz:
I think this issue supercedes #254, so I'm going to close that one.

aimalz:
To put this into context of other issues/ongoing branches, would it be a fair summary to say that it really pertains to the tests of whatever code we write to address #282, #285, #292, and #293?  If so, this issue need not correspond to a branch/PR of its own; we can instead plan to mark it with the "Don'tWorryWeGotThis" label when the branches for those issues (whose comments now show that they were linked here) are mature enough to have those tests written.

Issue #267, opened 2020-04-07
Bug in cluster_toolkit affecting modeling.py


marina-ricci:
I believe there are some bugs with the python wrapper of `cluster_toolkit`which propagate to CLMM/modeling.py. This includes silent crash, never ending computation or wrong results. So far, I noticed issue with the data type of the radius `r3d`. 

- when `r3d` is given as an 'int' instead of a float, no issue is raised but `get_3d_density` gives wrong result and `predict_excess_surface_density` crashes silently. This could be adressed by imposing `r3d` to be a float.

-  depending how `r3d` is computed,  `get_3d_density` may give wrong result and `predict_excess_surface_density` may crash silently. However, in python, I can not understand why it works or not as the data type of `r3d` is 'float64' in any case

This behaviour is illustrated in https://github.com/LSSTDESC/CLMM/blob/issue_267_bug_in_ct_and_modeling/examples/Bug_tester.ipynb

vitenti:
I just pushed a commit on this issue branch and I think it fixes the problem. Please check if it does not break something else.

combet:
@marina-ricci, @vitenti I don't think there has been a PR for this yet. If you think it's ready, please open one up so it may be reviewed and merged!

vitenti:
I just applied the fix directly in issue/282/modeling_structure (#282), I also copied the notebook that evinced the bugs. You can take a look there and close this and delete the branch.

combet:
OK - thanks

Issue #254, opened 2020-03-31
CLMM and NumCosmo comparison


vitenti:
In the comparison between CLMM modeling (using cluster toolkit) and NumCosmo, we found that the notebook modeling_cmp_numcosmo.ipynb gives suboptimal results. This happens for two different reasons:

- The value of the gravitational constant G in CLMM comes from astropy, thus, if you use astropy version < 4.0 it uses G from CODATA 2014 and if the version is >= 4.0 it uses G from CODATA 2018. Moreover, the value of the solar mass comes from the GSL (as it was in CCL, which was already updated). This relates with the issue #156 .

- The cor_factor was set to 1, whereas it should account for these differences in rho_critical. In addition there is cor_factor in wrong places.

We updated the notebook, fixing the above issues.

aimalz:
For completeness @m-aguena, do you think this is a continuation of #162?  In any case @vitenti, it sounds like this issue is ready for a Pull Request.

pennalima:
@marina-ricci, is the magnification going to be included in modeling? 
@aimalz and @m-aguena, I was planning to complete the magnification comparison before doing a pull request.

marina-ricci:
@pennalima Yes it will be included (if other agree on that). I am waiting for PR 264 and PR 266 before updating my PR because we are all working in `modeling.py`. Then I would need some help to implement the tests.

pennalima:
@marina-ricci Let me know when you implement the tests. I could help you.

vitenti:
For those interested, you can run NumCosmo + CLMM + Colossus online using the following link: 
https://mybinder.org/v2/gh/NumCosmo/NumCosmo/master

- The notebook NC_Colossus_density_profiles.ipynb shows the comparison between NumCosmo and Colossus.
- The notebook modeling_cmp_numcosmo.ipynb shows the comparison between NumCosmo and CLMM.

If that is something useful for any of you, I can prepare docker images containing NumCosmo + CLMM and others.

PR #299, opened 2020-08-02
Issue/282/modeling structure

vitenti:
This PR solves several issues:

- #144 cluster_toolkit is one of the possible backends, I don't think it should be included as a submodule/subproject since now we support other backends (CCL and NumCosmo) that are in active development. If all agree this should close that issue.
- #207 this issue concerns the different possible background mass definition (mean matter, critical, etc), this has out-of-the-box support in both CCL and NumCosmo. Therefore, this PR fixes this issue.
- #254 The comparison between NumCosmo and CLMM no longer makes sense since NumCosmo is now one of the possible backends. The comparison was in fact between cluster_toolkit and NumCosmo, there is a comparison notebook (CCL/Colossus/NumCosmo) in NumCosmo repo ( https://github.com/NumCosmo/NumCosmo/blob/master/notebooks/NC_CCL_Colossus_density_profiles.ipynb ).
- #267 Fixed by this PR.
- #277 Also fixed by this PR see #254 above.
- #282 The issue that originated this PR. The modeling module now is a script that loads from different backends for the functional interface that was developed in CLMM. It works transparently and load the backend depending on the value of a environment variable CLMM_MODELING_BACKEND (ct => cluster_toolkit, nc => numcosmo, ccl => ccl) . A OO interface was also included for CCL and NumCosmo. All modeling unit tests are now applied to all available backends. The actual backends are in the modbackend directory.
- #287 fixed by this PR.
- #296 also fixed here (also makes branch issue/254/clmm_numcosmo_cmp obsolete).


vitenti:
This PR requires NumCosmo version >= 0.15.2 to use its backend. (It is already available in conda-forge).

vitenti:
The drop on the coverage (coveralls) results from some backends (NumCosmo and CCL) not being available in the Coveralls build.

combet:
Thanks for the thorough PR @vitenti. Two quick questions/comments:
- I did not see any changes to the API documentation and noticed you exlcuded `api/clmm.modbackend.*` in `docs/conf.py`. Is that because these files are not yet documented?
- Regarding the drop in coverage, I understand it's because we're only installing cluster-toolkit in `.travis.yml` and not Numcosmo and CCL as well? Could we try adding a conda install for NC and CCL as well? I fear this will make the CI process a bit heavy though. What do you think?

combet:
Also, here're a few things/questions/comments, in no particular order (I'll update the list as I progress through the code)
-  in the OO approach, it would be useful to add a `backend` attribute to the `Modeling` object so that the user may easily check what backend they're using. e.g.
   - instantiating `moo = clmm.Modeling (massdef = 'mean', delta_mdef = 200, halo_profile_model = 'nfw')`,
   - `moo.backend` would then return 'nc' or 'ccl' depending on what has actually been used.

vitenti:
Hello @combet , I implemented your idea of having a property showing which backend is being used. Moreover, I just included CCL and NumCosmo to the travis build, it now takes about 5min, and still is a reasonable building time in my opinion. That fixed the coverage problem. 

About the documentation, we still need to decide the best way to document things (for instance, just one documentation in the abstract interface or one for each backend). That's why I removed these files from the doc building for the moment.

combet:
Thanks @vitenti for the latest modifications! Great to see Travis working with the 3 backends and coverage back up!

> About the documentation, we still need to decide the best way to document things (for instance, just one documentation in the abstract interface or one for each backend). That's why I removed these files from the doc building for the moment.

Thanks for clarifying. So I think this should be another issue to open once this PR is merged.

combet:
I also noticed the following in the OO modeling notebook:
- [ ] You have `DeltaSigma = moo.eval_sigma_mean (r3d, z_cl)`, but it should be DeltaSigma = moo.eval_sigma_excess (r3d, z_cl), right?

combet:
@vitenti, I ran into an issue when generating mock data with the `ccl` background in the "object oriented "`Example2..._oo.ipynb` (cell 7) that I added to this branch. I copy the end of the error message below. This occurs when asking for "photoz errors" on the mock data, i.e., with the line 

`noisy_data_z = mock.generate_galaxy_catalog(cluster_m, cluster_z, concentration, cosmo, ngals, Delta,'chang13', 
                                           shapenoise=0.05, 
                                           photoz_sigma_unscaled=0.05)`

Screenshot 2020-09-14 at 14 33 24

This runs fine with the `nc` backend. I will try to troubleshoot this further before this end of the week. Posting it here for reference, and any insights are very welcome !

vitenti:
Hello @combet it seems that the problem comes from the way things are vectorized in CCL, we must check if it is consistent with the way we are expecting things to be vectorized in CLMM. That's something that we should standardize.

vitenti:
For those interested I included two examples using NumCosmo statistical framework, you can find them in the branch issue/282/numcosmo_example the notebook Example2_Fit_Halo_Mass_to_Shear_Catalog_OO_NumCosmo.ipynb has a full MCMC analysis using both the wrong and correct approach. There I implemented the non-binned likelihood, I still didn't have time to implement the correct projected likelihood, but apart from the bias we can have an idea on how different will be the constraints using the two methods.

clmm/modbackend/ccl.py
```diff
@@ -0,0 +1,164 @@
+# Functions to model halo profiles
```
m-aguena, line 1: Can you add docstrings to the fucntions here? If the API here work exactly as in `clmm/modbackend/func_layer.py`, you could just copy them.

clmm/modbackend/numcosmo.py
```diff
@@ -0,0 +1,186 @@
+# NumCosmo implementation of CLMModeling
```
m-aguena, line 1: Can you add docstrings to the fucntions here? If the API here work exactly as in `clmm/modbackend/func_layer.py`, you could just copy them.

clmm/modbackend/func_layer.py
```diff
@@ -0,0 +1,509 @@
+# Thin functonal layer on top of the class implementation of CLMModeling .
```
m-aguena, line 1: Is there a reason for this and `generic.py` being separated files?

clmm/modbackend/ccl.py
```diff
@@ -0,0 +1,164 @@
...
+
+import pyccl as ccl
+
+from .cluster_toolkit_patches import _patch_comoving_coord_cluster_toolkit_rho_m
+
+import numpy as np
+import warnings
+
+from . import func_layer
+from . func_layer import *
```
m-aguena, line 11: please do not import `*`. Either import a module and use it or make specific imports

clmm/modbackend/numcosmo.py
```diff
@@ -0,0 +1,186 @@
...
+from gi.repository import NumCosmoMath as Ncm
+
+from .cluster_toolkit_patches import _patch_comoving_coord_cluster_toolkit_rho_m
+
+import math
+import numpy as np
+import warnings
+
+from . import func_layer
+from . func_layer import *
```
m-aguena, line 16: please do not import `*`. Either import a module and use it or make specific imports

clmm/modbackend/cluster_toolkit.py
```diff
@@ -0,0 +1,562 @@
+# Functions to model halo profiles
+
+import cluster_toolkit as ct
+import numpy as np
+from astropy import units
+from ..constants import Constants as const
+from .cluster_toolkit_patches import _patch_comoving_coord_cluster_toolkit_rho_m
+from . import generic
+from . generic import *
```
m-aguena, line 9: please do not import `*`. Either import a module and use it or make specific imports

clmm/modbackend/clmm_modeling.py
```diff
@@ -0,0 +1,36 @@
+# CLMModeling abstract class
```
m-aguena, line 1: If this parent class does not has any pre-set function, is it worth it to keep it? It just serve as an API template as it is.

clmm/modbackend/ccl.py
vitenti: In this file we have the OO interface to CCL, thus, it will inherit the docstring from CLMModeling, for this reason it is better to put all documentation in clmm_modeling.py and only add a docstring to the methods in ccl.py when we want to overwrite the father's documentation.

For example, lets say that in clmm_modeling.py we add a docstring documenting the behavior of set_mass, since CCLCLMModeling is a child of CLMModeling its implementation of set_mass will inherit the docstring from CLMModeling's set_mass method. If you want to see this working just use:

```
import inspect
print(inspect.getdoc(moo.set_mass))
```

and you will see the docstring that I just added to set_mass in clmm_modeling.py.

Since the documentation in func_layer.py concerns the functional interface they will be a bit different from the OO documentation. But here we have the same as in the oo interface, we just need to document the func_layer once and all back-ends will inherit it.

clmm/modbackend/numcosmo.py
vitenti: See above

clmm/modbackend/func_layer.py
vitenti: Yes, func_layer.py  is just a thin layer providing a functional interface for the OO CLMModeling object (implemented by an actual back-end). The file generic.py contains all functions that can be used by any back-end.

clmm/modbackend/ccl.py
vitenti: In general that's a good advice but in this case we are using the __all__ variable to control the exports from func_layer. That way, we do not need to repeat the functions names in all back-ends, and once a new function in added to func_layer __all__ variable, all back-ends will have it automatically.

clmm/modbackend/numcosmo.py
vitenti: See above.

clmm/modbackend/cluster_toolkit.py
vitenti: See above.

clmm/modbackend/clmm_modeling.py
vitenti: Using an abstract base class is good to control the subsequent implementations, take a look at [PEP 3119](https://www.python.org/dev/peps/pep-3119). Here we can put all documentation and common implementation of methods. We should actually make it better by using ABC in clmm_modeling.py .

aimalz: I think another way to phrase this is that we need it to support the back-ends just because they have modeling objects that behave differently, similarly to why we need a cosmology object (#293).

combet (approved):
It looks like @m-aguena will most likely be unavailable this week. @vitenti replied to all the points. Approving this PR. Will merge later in the week.

combet:
@vitenti  -  Regarding the docstrings, I saw the example you provided in `clmm_modeling` for `set_mass`, but the other functions are not documented yet. Do you agree we can open a separate issue and address this once this PR is merged? (We really need to get this PR through so we can move to other things that depend on it.)

vitenti:
@combet Yes, absolutely, we should merge ASAP.

PR #301, opened 2020-09-04
Issue/300/numcosmo ccl installdocs

cavestruz:


aimalz:
This is included in #299.

PR #302, opened 2020-09-21
Issue/284/improve mockdata

combet:
This addresses issue most of #284 (the rest is now under #303 and #304) and fixes the following:
- Check/improve documentation, in particular clarify comoving versus proper distance
- Fix a bug of 'true' versus 'observed' redshift: when redshift errors were added, the lensed quantities were computed using the observed `z` instead of the `ztrue`. This was a bug.
- Fixing the pdz to be centered on `zobs` and not `ztrue`, i.e. p(ztrue|zobs).
- Give the box size as an argument and not a hard coded value.
- Give the type of profile as an argument
- Give the option to pass a density of source galaxies, or a number of galaxies and a box size when generating the mock data.

The `demo_generate_mock_cluster` notebook has been updated to highlight some of the new f-o-v and galaxy density options.


combet:
@rherbonnet While updating the demo notebook for mock data generation, to highlight the new options we implemented in this branch, I came upon an issue with the new `ngal_density` option you implemented.
- at the moment, `ngals` is an argument of the `generate_galaxy_catalog` function, while `ngal_density` is a keyword. It means that `ngals` must be provided by the user, even if `ngal_density` is set to something (that will overwrites ngals).
- so I think we need to have `ngals` be a keyword as well and perform some checks in the function that return an error if:
  - neither `ngals` nor `ngal_density` are set
  - if both are set

Do you agree?

combet:
Also, when specifying a galaxy density (i.e. that we'd choose to mimick a given survey, e.g. 40 gal/arcmin2 for LSST), this number corresponds to the full depth of the survey, starting from z=0 and dropping to zero at high z. Therefore, when working with a redshift distribution (e..g, Chang et al. (2013)), the number of galaxies to draw needs to be renormalized to match the user-defined redshift range. I've modified the code to do that.

m-aguena:
NOTE:
Currently mock data returns `P(zphoto|ztrue)`, we will want to have `P(ztrue|zphot)`.
In this issue `P(ztrue|zphot)` will be implemented as a simple gaussian, another issue will be created for a more realistic PDF.

combet:
@m-aguena, following your advice I did merge master in my local branch before pushing again, but that did not solve the issue with one of the two Travis tests. The test suite is succeeding though (second Travis check). I'm not sure how to solve this. Any insight?

m-aguena:
I found a discussion similar to this error here:
https://github.com/travis-ci/travis-ci/issues/5850
I tried the solution but it did not work.
Maybe it will disappear once it is merged into master

combet:
Not sure how the Travis PR build got fixed, but now it passed...

examples/Example1_Fit_Halo_Mass_to_Shear_Catalog.ipynb
```diff
@@ -176,7 +176,7 @@
    "outputs": [],
    "source": [
     "ideal_data = mock.generate_galaxy_catalog(cluster_m, cluster_z, concentration,\n",
-    "                                          cosmo, ngals, Delta, src_z)"
+    "                                          cosmo, Delta, src_z, ngals=ngals)"
```
cavestruz, line 179: Relatively minor comment, but it would be good for the markdown at the top of this to describe the difference between this Example, Example2, and Example3.  Example 3 has a nice header.  For Example1, something like, **Fit Halo Mass from a Shear Catalog: The Ideal Data Example**

cavestruz (commented):
Some general comments - I like that ngas is specific in the cell of galaxy cluster parameters.

- [x] I think additional information in the notebook headers would be helpful.  It would be particularly useful to outline the differences between Example 1, Example 2, and Example 3 (both with specific header names, and bullet point list summary so it is easy to see the difference between each notebook).  Example suggestion for notebook 1: **Fit Halo Mass from a Shear Catalog: The Ideal Data Example**
- [x] Add descriptive Notebook Name headers that differentiate between each of the 3 Example?_Fit_Halo_Mass_to_Shear_Catalog
- [ ] Add bullet point summary of the notebook content
- [ ] I think we can remove all lines of `%matplotlib inline` (?)  I no longer use these in my locally jupyter notebooks, but this may not be the case on NERSC or IN2P3.

shenmingfu (commented):
It looks good to me. Thanks. My main question is about the normalization of the redshift distribution. Please see the comments on the code below.

examples/support/mock_data.py
```diff
@@ -120,13 +143,35 @@ def generate_galaxy_catalog(cluster_m, cluster_z, cluster_c, cosmo, ngals, Delta
     galaxy_catalog['id'] = np.arange(ngals)
     return galaxy_catalog
 
+def _chang_distrib(z):
+    """
+    A private function that returns the Chang et al (2013) galaxy redshift distribution function
+    """
+    alpha, beta, z0 = 1.24, 1.01, 0.51
+    return (z**alpha)*np.exp(-(z/z0)**beta)
```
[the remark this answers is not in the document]
combet: The docstrings now explicitly mention that this function is unnormalized. I'd rather limit docstrings to tell what the function is or does and not expend to what something it will be used for eslewhere in the code.

examples/support/mock_data.py
```diff
@@ -173,33 +216,54 @@ def _draw_source_redshifts(zsrc, cluster_z, zsrc_min, zsrc_max, ngals):
...
-
         def integrated_pzfxn(zmax, func):
             """Integrated redshift distribution function for transformation method"""
             val, _ = integrate.quad(func, zsrc_min, zmax)
             return val
         vectorization_integrated_pzfxn = np.vectorize(integrated_pzfxn)
 
         zsrc_domain = np.arange(zsrc_min, zsrc_max, 0.001)
-        probdist = vectorization_integrated_pzfxn(zsrc_domain, pzfxn)
+        probdist = vectorization_integrated_pzfxn(zsrc_domain, _chang_distrib)
```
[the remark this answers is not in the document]
combet: The transformation method is independent of the normalization I think. The line `uniform_deviate = np.random.uniform(probdist.min(), probdist.max(), ngals)` automatically accounts for it by uniformly sampling in `[probdist.min(), probdist.max()]`.

examples/Example1_Fit_Halo_Mass_to_Shear_Catalog.ipynb
combet: I think I've addressed all of your points above regarding make the differences between Example1-2-3 notebooks easier to grasp. Hope this is suitable.

marina-ricci (changes requested):
Nice improvements! 
Minor suggested modifications : 

In the docstring of `mock_data.py`: 

- [x] specify that the requested zmin and zmax will be applied on the *true* redshift distribution. 
- [x] modify in the item 3.  "a side length of 8 Mpc" > "a side length of 8 Mpc/h."

In `demo_generate_mock_data`

- [x] you could add lines to indicate the obsreved and true redshift in teh plot where you show the "pdf".

combet:
@marina-ricci, thank you for the review. All your points have been addressed.

marina-ricci (approved):
Great, thank you!

examples/support/mock_data.py
```diff
@@ -78,15 +77,27 @@ def generate_galaxy_catalog(cluster_m, cluster_z, cluster_c, cosmo, ngals, Delta
         str : Draws individual source gal redshifts from predefined distribution. Options
               are: chang13
     zsrc_min : float, optional
-        The minimum source redshift allowed.
+        The minimum true redshift of the sources. If photoz errors are included, the observed redshift
+        may be smaller than zsrc_min.
     zsrc_max : float, optional
-        If source redshifts are drawn, the maximum source redshift
+        The maximum true redshift of the sources, apllied when galaxy redshifts are drawn from 
+        a redshift distribution. If photoz errirs are included, the observed redshift may be larger than
```
shenmingfu, line 84: A small typo here. errirs -> errors

examples/support/mock_data.py
```diff
@@ -163,39 +225,60 @@ def _generate_galaxy_catalog(cluster_m, cluster_z, cluster_c, cosmo, ngals, Delt
...
+              are: chang13
+    zsrc_min : float
+        The minimum source redshift allowed.
+    zsrc_max : float, optional
+        If source redshifts are drawn, the maximum source redshift
+
+    Returns
+    -------
+    galaxy_catalog : clmm.GCData
+        Table of true and 'measured' redshifts, which here the same. Photometric errors
```
shenmingfu, line 252: It would be better to add "redshift" between "photometric" and "errors".

examples/support/mock_data.py
```diff
@@ -204,24 +287,39 @@ def integrated_pzfxn(zmax, func):
...
+    """
     galaxy_catalog['pzsigma'] = photoz_sigma_unscaled*(1.+galaxy_catalog['ztrue'])
     galaxy_catalog['z'] = galaxy_catalog['ztrue'] + \
-                          galaxy_catalog['pzsigma']*np.random.standard_normal(ngals)
+                          galaxy_catalog['pzsigma']*np.random.standard_normal(len(galaxy_catalog))
+
     pzbins_grid, pzpdf_grid = [], []
     for row in galaxy_catalog:
-        zmin, zmax = row['ztrue'] - 0.5, row['ztrue'] + 0.5
+        zmin, zmax = row['z'] - 0.5, row['z'] + 0.5
```
shenmingfu, line 312: This would cause the (redshift) interval (for the probability) to be different for different galaxies. But maybe it's okay.

shenmingfu (commented):
Thank you. Sorry I just found this had been merged. I noticed some small issues. They are not urgent and can be fixed in the future.

combet:
> Thank you. Sorry I just found this had been merged. I noticed some small issues. They are not urgent and can be fixed in the future.

Thanks for pointing those out. I hadn't realised you still had some points to address and merged once Marina approved the PR. I'll implement those small corrections in #306, which is the next thing that will need to be reviewed and merged.

examples/support/mock_data.py
combet: I confirm this is not a problem.

examples/support/mock_data.py
combet: Will correct this in #306

examples/support/mock_data.py
combet: Will correct this in #306

Issue #308, opened 2020-10-09
Typo in gcdata.py line 57


hsinfan1996:
Line 57: for name, vale in zip(names, values), "vale" should be changed to "value" because "value" is the variable taken in the next line.

PR #315, opened 2020-10-26
Issue/313/add global optimization to sampler.py

hsinfan1996:
Added scipy.optimize.basinhopping as a new method to compute the minimum, which is more reliable than scipy.optimize.minimum.
Fixed a typo in gcdata.py line 57 (issue #308)
There are some small changes to Example1. Now random seed is set to 11 (again), which seems to have been used when the example was written. samplers['minimize'] is replaced by samplers['basinhopping'] in order to have consistent results of  logm_est1 and logm_est2 regardless of the random seeds.


combet (changes requested):
Thanks @hsinfan1996 , this looks good! 

May I just ask that you keep the `sampler['minimize']` code in a cell, but commented out, along with a new markdown cell explaining that for some random seeds it has proven unstable, a limitation that the `basinhopping` method does not seem to have. That way, the user still has an example of the code to call to work with `sampler['minimize']`, if they want to.

hsinfan1996:
In our use case, basinhopping method does not seem to have limitations. I only encountered errors when trying to calculate estimate masses with thousands of different initial guesses.

hsinfan1996:
By the way, I think I just spotted another typo in the last cell in the Example1 notebook. 
Screen Shot 2020-11-10 at 16 36 02
best fit model 2 should plot gt_model2 instead of gt_model1

hsinfan1996:
The commits look messy because I was trying to solve a conflict from gcdata.py. Essentially, I only made changes to Example1. Others were pulled from the latest master.

hsinfan1996:
This PR has become so messy. Hope it can be merged soon.

Issue #312, opened 2020-10-19
Remaining things to improve/correct in mock_data


rherbonnet:
- [x] In mock_data._find_aphysical_galaxies there's the lines:
`    badgals = np.where((np.abs(galaxy_catalog['e1']) > 1.0) |
                       (np.abs(galaxy_catalog['e2']) > 1.0) |
                       (galaxy_catalog['ztrue'] < zsrc_min)
                      )[0] `
So all galaxies with either |e1|>1 or |e2|>1 are rejected. But shouldn't it be rejecting |e|=sqrt(e1 x e1+e2 x e2)>1, instead?

- [x] In `_generate_galaxy_catalog`, there remain some hard-coded `halo_profile_model='nfw'` that should be changed to `halo_profile_model=halo_profile_model` as this is a user-defined keyword. With the Numcosmo backend, other profiles are available.

Once #311 is merged:

- [x] Suggestion to add another keyword `massdef` that can be equal to `mean`, `critical` or `virial`. At the moment, mock_data can only operate with the mean overdensity mass definition. With the CCL and NC backend, the user could work with the other definitions and the option should be available in mock_data.  

- [x] Also, Delta_SO is a non-optional argument of `generate_galaxy_catalog`. Suggest making it an optional keyword, that defaults to 200, for consistency with `massdef` being a optional keyword as well.

combet:
@rherbonnet, I hijacked your issue to add 2 more things that we should take care of in mock_data. Didn't want to open a separate issue as they all concern the same file.

shenmingfu:
In Example 1, binning would cause fitted mass larger than input. The reason is that g(r) is a convex function -- the function value after binning would be larger, but the bias becomes smaller as bin number increases. I will try to add some description to Example1 to highlight this.

PR #316, opened 2020-10-26
Issue/312/mock data extra

combet:
Adresses #312.
- In mock_data, all galaxies with either |e1|>1 or |e2|>1 were rejected. Now, only reject those with |e|=sqrt(e1 x e1+e2 x e2)>1.
- Remove remaining some hard-coded `halo_profile_model='nfw'`. Changed to halo_profile_model=halo_profile_model as this is a user-defined keyword. With the Numcosmo backend, other profiles are available.

In `generate_galaxy_catalog`:
- Add a `massdef` optional keyword so the user may generate mock data with the other definitions and the option should be available in mock_data.
- Change `Delta_SO` to be an **optional** keyword (default to 200), for consistency with `massdef`.

Notebooks:
- Extand `demo_generate_mock_cluster` to show usage with the new options
- Update all notebooks as `Delta` is now an optional keyword with a default value.


combet:
@vitenti , @pennalima  - I've updated the `demo_generate_mock_cluster` notebook for this PR, to highlight the use of the new user options that become available with the NC and/or CCL backend ('mean' or 'critical', 'nfw' or 'einasto', etc.).

However, trying to run the notebook with the CCL backend, an error was raised when some source are in front of the lens (we had that already in the past and I'm pretty sure you fixed it already), but I've traced it back to `ccl.py` where the cut on the 'good' lens-src pairs was not propagate to `a_len`. I think it's OK now, but I would appreciate if you could take a quick look.

clmm/modbackend/ccl.py
```diff
@@ -97,8 +97,8 @@ def eval_sigma_crit(self, z_len, z_src):
 
         if np.any(z_cut):
             Ds = ccl.angular_diameter_distance(self.cosmo.be_cosmo, a_src[z_cut])
-            Dl = ccl.angular_diameter_distance(self.cosmo.be_cosmo, a_len)
-            Dls = ccl.angular_diameter_distance(self.cosmo.be_cosmo, a_len, a_src[z_cut])
+            Dl = ccl.angular_diameter_distance(self.cosmo.be_cosmo, a_len[z_cut])
+            Dls = ccl.angular_diameter_distance(self.cosmo.be_cosmo, a_len[z_cut], a_src[z_cut])
```
shenmingfu, line 101: This requires the lengths of a_len and a_src are equal. Can we double check that?

examples/support/mock_data.py
```diff
@@ -65,20 +65,26 @@ def generate_galaxy_catalog(cluster_m, cluster_z, cluster_c, cosmo, Delta_SO, zs
...
     zsrc : float or str
         Choose the source galaxy distribution to be fixed or drawn from a predefined distribution.
         float : All sources galaxies at this fixed redshift
         str : Draws individual source gal redshifts from predefined distribution. Options
               are: chang13
+    Delta_SO : float, optional
+        Overdensity density contrast used to compute the cluster mass and concentration. The
+        spherical overdensity mass is computed as the mass enclosed within the radius
+        :math:`R_{\Delta{\rm SO}}` where the mean density is :math:`\Delta_{\rm SO}` times
+        the mean density of the Universe at the cluster redshift
```
shenmingfu, line 77: Maybe it's better to add parentheses: the mean (matter) density?

shenmingfu (commented):
This looks great to me. Thank you. I added some comments, please take a look. Thanks.

examples/support/mock_data.py
combet: Thanks, done.

clmm/modbackend/ccl.py
combet: @shenmingfu Thanks for pointing this potential source of error. I looked into it but actually, the way the modeling functions are vectorized do not support z_cl to be array-like. It should be a float and the docstrings in `clmm_modeling.py` specify that. So I don't think we need to perform an extra check here as `a_len` should be a float, which is then converted into an array of the same length as `a_src` by `a_len = np.repeat(a_len, len(a_src))`.

So I think this is OK and hope this answers your concern.

shenmingfu (approved):
Thank you very much for answering my questions and adjusting the code.

PR #317, opened 2020-10-28
Issue/314/cosmology consistency

m-aguena:
Functions to check for cosmology consistency were added to the Cosmology object:

- `update_cosmo_ext_valid`: checks the input cosmology with an external GCData
- `update_cosmo`: checks the input cosmology with internal cosmology

And they are being used on `compute_tangential_and_cross_components` and `make_binned_profile` in the `polaraveraging` module.


vitenti (approved):
My only concern with this solution is that since it compares the string representations of cosmologies to assert that they are the same, it can fail even if the two cosmologies are the same but one parameter differs by a machine epsilon (~1.0e-16). For instance, the string representaton of 0.3 is sometimes 0.300000000000001 due to rounding errors. However, it is a reasonable enough solution for the time being.

m-aguena:
> My only concern with this solution is that since it compares the string representations of cosmologies to assert that they are the same, it can fail even if the two cosmologies are the same but one parameter differs by a machine epsilon (~1.0e-16). For instance, the string representaton of 0.3 is sometimes 0.300000000000001 due to rounding errors. However, it is a reasonable enough solution for the time being.

I created issue #322 to solve this later.

combet (commented):
Thanks @m-aguena! Before merging, I tried this out, by adapting an example from the notebook and it works as advertised :). 

However, as they are, none of the example notebooks allow to directly test it: the example GCData galaxy catalogs (generated either from `mock_data.py` or loaded from the DC2 data) do not set the cosmology metadata and `galcat.meta['cosmo']` is `None`. This is super easy to change, adding one line using your new `update_cosmo(cosmo, overwrite=True)` function, either in `mock_data.py` or in the 2 DC2 example notebooks.

I think it would be good to have the examples updated either in this PR or in a separate issue. If you think the latter is best, let me know and I'll merge this PR straight away (allowing to move to the next one ;) )

m-aguena:
> Thanks @m-aguena! Before merging, I tried this out, by adapting an example from the notebook and it works as advertised :).
> 
> However, as they are, none of the example notebooks allow to directly test it: the example GCData galaxy catalogs (generated either from `mock_data.py` or loaded from the DC2 data) do not set the cosmology metadata and `galcat.meta['cosmo']` is `None`. This is super easy to change, adding one line using your new `update_cosmo(cosmo, overwrite=True)` function, either in `mock_data.py` or in the 2 DC2 example notebooks.
> 
> I think it would be good to have the examples updated either in this PR or in a separate issue. If you think the latter is best, let me know and I'll merge this PR straight away (allowing us to move to the next one ;) )

Yes, sorry, I only remembered to add some examples of it when working on #320. It is on `examples/demo_dataops_functionality.ipynb` there. I suggest, if there are not any other comment, we accept this PR as it is and move this discussion to PR #321.

combet:
Ah, OK great - thanks @m-aguena. Will proceed with the merge then and will move to review #321.

PR #321, opened 2020-11-04
Issue/320/rm galaxycluster from dataops

m-aguena:
Main changes:

- Renamed `polaraveraging.py` -> `dataops.py`
- Renamed function `make_binned_profile` -> `make_radial_profile` 
- Removed cluster from input of `compute_tangential_and_cross_components` and `make_radial_profile`
- Added `compute_tangential_and_cross_components` and `make_transversal_profile` to `GalaxyCluster`
- Created `add_critical_surface_density` function to add sigma_crit in `GalaxyCluster.galcat`

Note:

PR #317 should be merged before this one.


m-aguena:
Just a reminder to check for this issue raised by @combet in this PR:
https://github.com/LSSTDESC/CLMM/pull/317#pullrequestreview-526150584

(it might already be resolved in `examples/demo_dataops_functionality.ipynb`)

combet:
Hi @m-aguena - so I merged #317 as requested to start the review of this PR but it is now in conflict with `master`... Could you please take a look?

clmm/dataops.py
```diff
@@ -0,0 +1,275 @@
...
+#     import pyccl as ccl
+# except ImportError:
+#     pass
+import math
+import warnings
+import numpy as np
+from .gcdata import GCData
+from .utils import compute_radial_averages, make_bins, convert_units
+from .modeling import get_critical_surface_density
+def compute_tangential_and_cross_components(
```
combet, line 12: I noticed there are no blank lines above any function definitions in `dataops.py` or in the methods of `galaxycluster.py`. Is it due to some autoformatting at your end? I find it  makes the code hard to read. Also I think PEP8 recommands two blank kines between function definitions.

clmm/dataops.py
```diff
@@ -0,0 +1,275 @@
...
+    .. math::
+        g_x = g_1 \sin\left(2\phi\right)-g_2\cos\left(2\phi\right)
+
+    For extended descriptions of parameters, see `compute_shear()` documentation.
+    """
+    return shear1*np.sin(2.*phi)-shear2*np.cos(2.*phi)
+def make_transversal_profile(components, angsep, angsep_units, bin_units,
+                        bins=10, include_empty_bins=False,
+                        return_binnumber=False,
+                        cosmo=None, z_lens=None):
```
combet, line 190: This used to be called `make_binned_profile` and I am not sure I understand the need for a different name. Could you please clarify your thoughts there. I find the adjective `transversal` quite confusing and I don't think it is used in the literature. I'd suggest `radial` instead and maybe `make_binned_radial_profile` for the function name, which is very explicit (but maybe too long). Pleas share your thoughts :) .

combet (commented):
Thanks @m-aguena, this is very neat and makes this part of the code sooo much clearer. I have a few comments above but also noticed:
- in the `demo_dataops` notebook, the Note at the top still mentions the possibility to use `compute_tangential_and_cross_components(cl, geometry="flat")`, which is not the case anymore
- one of the Travis run failed, but I believe this is at Travis's end.

I haven't gone through everything yet and may yet add some more comments.

clmm/modbackend/numcosmo.py
```diff
@@ -228,7 +228,7 @@ def _get_param(self, key):
 
     def set_dist(self, dist):
 
-        assert isinstance(dist, Nc.Distance)
+        #assert isinstance(dist, Nc.Distance)
```
combet, line 231: Just curious why this is not needed anymore...

clmm/dataops.py
m-aguena: no problem, I will change it

clmm/dataops.py
m-aguena: @combet I was not very happy with the name `make_binned_profile`, it sounded redundant and not explicit enough at the same time. I added the word `transversal` because that is how we were binning, in a transversal distance (be it angular or physical). But I understand it is not widely used and it might generate confusion. How about `make_radial_profile`?

clmm/modbackend/numcosmo.py
m-aguena: this change should not have been committed. Good thing you caught it :wink:

m-aguena:
> Thanks @m-aguena, this is very neat and makes this part of the code sooo much clearer! I have a few comments above but also noticed:
> 
> * in the `demo_dataops` notebook, the Note at the top still mentions the possibility to use `compute_tangential_and_cross_components(cl, geometry="flat")`, which is not the case anymore
> * one of the Travis run failed, but I believe this is at Travis's end.
> 
> I haven't gone through everything yet and may yet add some more comments.

I updated this notebook

clmm/dataops.py
combet: @m-aguena , yes, `make_radial_profile` would work well, with the radial distance that can be understood as angular or physical distance.

combet:
@m-aguena, some more things I noticed when running the `dataops` notebook:
- **Cell 9** (not using `compute_tangential_and_cross_components` as a method of the cluster object): As it is written, `theta`, `e_t` or `g_x` (which should be `e_x`, typo to fix) are Astropy columns (i.e. the same type as `cl.galcat['theta',...]`) and if you look at them, they are named `ra`, `e1` and `e1`. That's something we noticed before and the quick fix (that you can see in `polaraveraging` in the `master` branch), was to cast the results as `np.array` before returning (so that no name would be attached to them).

In `dataops.py`, that would mean changing line 130 to`return np.array(angsep), np.array(tangential_comp), np.array(cross_comp)`

Alternatively, the astropy columns can be cast as arrays in the function call of Cell 9, so that the outputs are also simplu numpy arrays.
```
theta, e_t, e_x = compute_tangential_and_cross_components(
                ra_lens=cl.ra, dec_lens=cl.dec,
                ra_source=np.array(cl.galcat['ra']), dec_source=np.array(cl.galcat['dec']),
                shear1=np.array(cl.galcat['e1']), shear2=np.array(cl.galcat['e2']),
                geometry="flat")
```
There might be a better way, but we should prevent this random naming of the output when the input are `Columns` and not arrays.


- In **Cell 13** (again when not using the function as a method of the cluster objects), the GCData output has columns named `p_i, p_i_err`. While this is explained in the doctstrings of the `make_***_profile` function, I think it would be useful to have a sentence in the notebook explaining what they are.

- Finally for the cells that exemplify using the functions as method of the cluster object, would it better to simply have 
`cl.compute_tangential_and_cross_components(geometry="flat", add=True)` 
instead of 
`theta, e_t, g_x = cl.compute_tangential_and_cross_components(geometry="flat", add=True)`? We can but don't need to store the results in new variables, as the table of the cluster object will contain the result. Just a question, no strong opinion.

clmm/dataops.py
m-aguena: done!

m-aguena:
> @m-aguena, thanks for the changes above. Some more things I noticed when running the `dataops` notebook:
> 
> * **Cell 9** (not using `compute_tangential_and_cross_components` as a method of the cluster object): As it is written, `theta`, `e_t` or `g_x` (which should be `e_x`, typo to fix) are Astropy columns (i.e. the same type as `cl.galcat['theta',...]`) and if you look at them, they are named `ra`, `e1` and `e1`. That's something we noticed before and the quick fix (that you can see in `polaraveraging` in the `master` branch), was to cast the results as `np.array` before returning (so that no name would be attached to them).
> 
> In `dataops.py`, that would mean changing line 130 to`return np.array(angsep), np.array(tangential_comp), np.array(cross_comp)`
> 
> Alternatively, the astropy columns can be cast as arrays in the function call of Cell 9, so that the outputs are also simply numpy arrays.
> 
> ```
> theta, e_t, e_x = compute_tangential_and_cross_components(
>                 ra_lens=cl.ra, dec_lens=cl.dec,
>                 ra_source=np.array(cl.galcat['ra']), dec_source=np.array(cl.galcat['dec']),
>                 shear1=np.array(cl.galcat['e1']), shear2=np.array(cl.galcat['e2']),
>                 geometry="flat")
> ```
> 
> There might be a better way, but we should prevent this random naming of the outputs when the inputs are `Columns` and not arrays.
> 
> * In **Cell 13** (again when not using the function as a method of the cluster objects), the GCData output has columns named `p_i, p_i_err`. While this is explained in the doctstrings of the `make_***_profile` function, I think it would be useful to have a sentence in the notebook explaining what they are.
> * Finally for the cells that exemplify using the functions as method of the cluster object, would it better to simply have
>   `cl.compute_tangential_and_cross_components(geometry="flat", add=True)`
>   instead of
>   `theta, e_t, g_x = cl.compute_tangential_and_cross_components(geometry="flat", add=True)`? We can but don't need to store the results in new variables, as the table of the cluster object will contain the result. Just a question, no strong opinion.

 - **Cell 9:** I fixed the typo, and updated input check  of `compute_tangential_and_cross_components` with a function `arguments_consistency` that also fix the name of the outputs.
 - **Cell 13:** I added a sentence right after it explaining the table columns.
 - **Final:** I removed the use of getting outputs from `GalaxyCluster` functions (ex: `theta, e_t, g_x = cl.compute_tangential_and...`)

combet (approved):
Thanks for the changes @m-aguena. All looks good! Approving this PR.

PR #328, opened 2020-11-24
Add gha support

vitenti:
Adding support for GitHub actions, once this is stable and reviewed Travis-ci support can be dropped. The documentation generating code (had to be) was updated and works with the latest sphinx version available. 

- [ ] Check if the documentation is being built correctly.
- [ ] Check if coverage upload is correct.
- [ ] Check if it is necessary to change branch configuration in order to run GHA in every branch.

### DIFF
--- a/.github/workflows/build_check.yml
+++ b/.github/workflows/build_check.yml
@@ -14,7 +14,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Install prereq using apt
-      run: sudo apt-get install libgsl0-dev swig3.0 libfftw3-dev gobject-introspection python3-gobject gfortran libmpfr-dev libhdf5-dev liblapack-dev libnlopt-dev libcfitsio-dev gir1.2-glib-2.0 libgirepository1.0-dev python3-gi python3-numpy python3-matplotlib python3-setuptools pandoc -y
+      run: sudo apt-get install libgsl0-dev swig3.0 libfftw3-dev gobject-introspection python-gobject gfortran libmpfr-dev libhdf5-dev liblapack-dev libnlopt-dev libcfitsio-dev gir1.2-glib-2.0 libgirepository1.0-dev python3-gi python3-numpy python3-matplotlib python3-setuptools pandoc -y
     - name: Install cluster_toolkit from source
       run: |
         git clone https://github.com/tmcclintock/cluster_toolkit.git
--- a/.github/workflows/build_check.yml
+++ b/.github/workflows/build_check.yml
@@ -42,6 +42,9 @@ jobs:
     - name: Run the unit tests
       run: |
         pytest-3 tests/ --ignore=cluster_toolkit/tests --cov=clmm/
+    - name: Install Sphinx prereq
+      run: |
+        sudo pip3 install sphinx_rtd_theme
     - name: Run the docs
       run: |
         make -C docs/ html
--- a/.github/workflows/build_check.yml
+++ b/.github/workflows/build_check.yml
@@ -48,7 +48,8 @@ jobs:
     - name: Run the docs
       run: |
         make -C docs/ html
-    - name: Coveralls Finished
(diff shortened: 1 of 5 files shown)


vitenti:
I had to commit several times until I had everything working correctly, so when merging remember to squash everything in a single commit.

combet:
Thanks @vitenti for GHA support, it's great! There's just one thing I'd like to understand - the `master` branch page (snapshot below) now suggests that the Travis build is not passing (which it was). But actually, clicking on the red cross brings us to a failed Travis build of the `add_gha` branch. So I guess `master` is still fine, but I'd like to understand where that mix up comes from.

Screenshot 2020-11-24 at 09 22 30

vitenti:
I was searching for the branch add_gha_support in the GitHub branch search box and end-up pressing enter before finishing writing, this created a new branch called add_gha, I deleted it but I think that confused Travis.

vitenti:
If you look at https://travis-ci.org/github/LSSTDESC/CLMM/branches you can see that the master branch was built correctly.

johannct:
It looks good to me. I have two questions/comments, none blocking the merge : 
- `on: [push, pull_request]` : this means that you run on every single branch. It is ok, but in RAIL we found that a bit overkill and went for 
`on:
  push:
    branches: [ $default-branch ]
  pull_request:
    branches: [ master ]
`
so that especially for pull requests only pulls into master are concerned.
- I do not have religion about coverage webapps. I just noticed that CLMM followed CCL in using coveralls while BlendingToolkit, ceci, RAIL and Coord use code-cov.

combet:
Thanks @johannct for checking this out and for sharing what is used in RAIL. Just to make sure I understand, in the configuration you're suggesting, if I push something into a branch that is not `master` and that is not part of a PR to be merged into `master`, no build would be triggered?

@vitenti, I looked the documentation and didn't see anything amiss (we have some formatting issues, but they were there before and will be taken care of in a specific issue). I have 2 remaining questions:
- You suggested dropping Travis support once gha is "stable and reviewed". So do you want to do this simultaneously with this PR or wait a bit longer after this PR is merged and do this in a subsequent step? 
- Could you update the 'build passing' badge to point to the gha output instead of the Travis one?

johannct:
yes, one can (and should!) always run the test suite by hand during development. But perhaps the triggering of the GHA-based CI does not need to happen at every single push into any branch. Likewise for coverage studies.

vitenti:
I agree that running coverage on every push and branch is unnecessary. In my experience, running CI on every push and branch can be useful. Sometimes there is a particular configuration on the developer machine that hides a bug, so running in a freshly installed version (which is what a CI does) can help find these bugs.

johannct:
You would get this kind of bug on the pull requests. And pull requests should not linger or concern huge amounts of codes, so it would happen early enough to not add a debugging burden (agile principle)

vitenti:
@combet Maybe we could let Travis-ci as it is now but no longer gives support to it (if at any point in the future Travis-ci's build breaks we just remove Travis-ci support). 

@johannct So, since we do not push into master directly, CI/coverage would be run only on PR's, right? What would be the advantage of doing that (or the downside of not doing it)?

vitenti:
Badge updated!

johannct:
ultimately it is a matter of taste. If the CLMM team likes it that way, it is fine with me, and we can always modify it later.

combet:
> Badge updated!

Great, thank you :) 

> if at any point in the future Travis-ci's build breaks we just remove Travis-ci support. 

From [this page](https://docs.travis-ci.com/user/migrate/open-source-repository-migration/#frequently-asked-questions), it seems`travis-ci.org` will be deprecated and become read-only at the end of the year (projects are supposed to migrate to `travis-ci.com` instead), so I guess we'll see it break and will remove it then.

I'll go ahead and approve the PR. We can always revisit CI parameters as needed in the future.

combet (approved):
Everything looks good!

PR #331, opened 2020-11-27
Issue/330/make_ct_oo

m-aguena:
Add object oriented layer to `cluster_toolkit`. Additional changes:
 - Make CLMMModeling convert different CLMM cosmology objects to the required one.
 - Add unit tests for objected oriented approaches.


vitenti (approved):
Everything seems well thought out!

combet (changes requested):
Thanks @m-aguena! I tried running `pytest` on my laptop where I have CT and NC installed but not CCL. And it failed because the tests now tries to import `pyccl`, even though it is not available. I didn't identify exactly where it comes from, but needs to be fixed. The tests ran fine on an install where I have all backends available (would have missed it running only there).

```
ImportError while loading conftest '/Users/combet/RECHERCHE/LSST/CLMM/tests/conftest.py'.
tests/conftest.py:4: in <module>
    import clmm
clmm/__init__.py:3: in <module>
    from .galaxycluster import GalaxyCluster
clmm/galaxycluster.py:7: in <module>
    from .dataops import compute_tangential_and_cross_components, make_radial_profile
clmm/dataops.py:11: in <module>
    from .modeling import get_critical_surface_density
clmm/modeling.py:80: in <module>
    backend = importlib.import_module(".modbackend."+__backends[be_nick]['module'], package=base_package)
clmm/modbackend/cluster_toolkit.py:18: in <module>
    from .ccl import CCLCosmology
clmm/modbackend/ccl.py:3: in <module>
    import pyccl as ccl
E   ModuleNotFoundError: No module named 'pyccl'
```

clmm/modbackend/func_layer.py
```diff
@@ -39,13 +39,13 @@ def get_3d_density(r3d, mdelta, cdelta, z_cl, cosmo, delta_mdef=200, halo_profil
     halo_profile_model : str, optional
         Profile model parameterization, with the following supported options:
             `nfw` (default)
-            `einasto`
-            `hernquist`
+            `einasto` (not in cluster_toolkit)
+            `hernquist` (not in cluster_toolkit)
```
combet, line 43: Actually, running the OO modeling demo with `ccl` and `einasto`, returns a CCL error, linked I think to the numerical integration in CCL (I remember @pennalima metionning there was a problem in CCL along those lines). So really, `einasto` is not supported by the CCL backend either.

Regarding the `hernquist` profile with CCL, the demo notebook runs but I saw significant numerical instabilities in the plots, so I'm not sure it is properly supported either. Maybe @pennalima or @vitenti  can confirm?

clmm/modbackend/cluster_toolkit.py
```diff
@@ -1,24 +1,24 @@
...
-from . generic import get_reduced_shear_from_convergence
+from . import func_layer
+from . func_layer import *
 
 from .. clmm_cosmo import CLMMCosmology
+from .. clmm_modeling import CLMModeling
 
-import warnings
+from .ccl import CCLCosmology
+from .numcosmo import NumCosmoCosmology
```
combet, line 19: Not sure I understand while these need to be imported `cluster_toolkit` module.

combet:
@m-aguena Oh, and one last thing... The OO notebooks (demo_modeling, Example2...) mention that they will run only with `nc` and `ccl`. This should be updated to mention that they'll run with `ct` as well.

combet (commented):
Oh, and one last thing... The OO notebooks (demo_modeling, Example2...) mention that they will run only with `nc` and `ccl`. This should be updated to mention that they'll run with `ct` as well.

vitenti:
> Thanks @m-aguena! I tried running `pytest` on my laptop where I have CT and NC installed but not CCL. And it failed because the tests now try to import `pyccl`, even though it is not available. I didn't identify exactly where it comes from but needs to be fixed. The tests ran fine on an install where I have all backends available (would have missed it running only there).
> 
> ```
> ImportError while loading conftest '/Users/combet/RECHERCHE/LSST/CLMM/tests/conftest.py'.
> tests/conftest.py:4: in <module>
>     import clmm
> clmm/__init__.py:3: in <module>
>     from .galaxycluster import GalaxyCluster
> clmm/galaxycluster.py:7: in <module>
>     from .dataops import compute_tangential_and_cross_components, make_radial_profile
> clmm/dataops.py:11: in <module>
>     from .modeling import get_critical_surface_density
> clmm/modeling.py:80: in <module>
>     backend = importlib.import_module(".modbackend."+__backends[be_nick]['module'], package=base_package)
> clmm/modbackend/cluster_toolkit.py:18: in <module>
>     from .ccl import CCLCosmology
> clmm/modbackend/ccl.py:3: in <module>
>     import pyccl as ccl
> E   ModuleNotFoundError: No module named 'pyccl'
> ```

That's a type of bug hard to catch using pytest. The problem is that I have all backends installed in my machine, so I can't see these bugs. I tried to implement some way to hide available modules from python (tweaking the import system), to check if tests run when some backends are not available, but it does not work for any python version so I abandoned it. We should find a solution to emulate different settings so that the unit test could catch these bugs...

clmm/modbackend/func_layer.py
vitenti: Yes, their default settings give very poor results, I would mark anything other than NFW as 'unstable' under the CCL backend.

clmm/modbackend/cluster_toolkit.py
vitenti: @m-aguena added support for converting between different cosmologies. So, these modules are imported because it checks if the cosmology is "convertible" looking at the class it comes from (so it needs to import CCL and NumCosmo). 

I think that this feature can be a bit dangerous. The conversion does not guarantee that the output cosmology will be exactly the same (we would have to mess with the other aspects of the background cosmologies, such as neutrino masses, Yb, reionization redshift, etc). Thus, the way it is now if you pass a NumCosmo cosmology to a CT object it will just silently convert by using the same values of the Omegas to create an AstroPy cosmology object but will ignore other cosmological parameters. If we are going to keep this feature I suggest adding a big warning every time such conversion happens.

m-aguena: these two were imported so `CTModeling.set_cosmo` could check on the types of cosmology to be accepted. But you are right, it does impose to have `ccl` installed. I will implement a better way to check it without having those imports.

m-aguena:
> > Thanks @m-aguena! I tried running `pytest` on my laptop where I have CT and NC installed but not CCL. And it failed because the tests now try to import `pyccl`, even though it is not available. I didn't identify exactly where it comes from but needs to be fixed. The tests ran fine on an install where I have all backends available (would have missed it running only there).
> > ```
> > ImportError while loading conftest '/Users/combet/RECHERCHE/LSST/CLMM/tests/conftest.py'.
> > tests/conftest.py:4: in <module>
> >     import clmm
> > clmm/__init__.py:3: in <module>
> >     from .galaxycluster import GalaxyCluster
> > clmm/galaxycluster.py:7: in <module>
> >     from .dataops import compute_tangential_and_cross_components, make_radial_profile
> > clmm/dataops.py:11: in <module>
> >     from .modeling import get_critical_surface_density
> > clmm/modeling.py:80: in <module>
> >     backend = importlib.import_module(".modbackend."+__backends[be_nick]['module'], package=base_package)
> > clmm/modbackend/cluster_toolkit.py:18: in <module>
> >     from .ccl import CCLCosmology
> > clmm/modbackend/ccl.py:3: in <module>
> >     import pyccl as ccl
> > E   ModuleNotFoundError: No module named 'pyccl'
> > ```
> 
> That's a type of bug hard to catch using pytest. The problem is that I have all backends installed in my machine, so I can't see these bugs. I tried to implement some way to hide available modules from python (tweaking the import system), to check if tests run when some backends are not available, but it does not work for any python version so I abandoned it. We should find a solution to emulate different settings so that the unit test could catch these bugs...

@vitenti @combet one way of working around the failures of not having all backends installed it to put the tests for each backend on a different file. This way, when testing locally, you can just comment the files for backends you don't want to test for. I can make this modification.

combet:
> @vitenti @combet one way of working around the failures of not having all backends installed it to put the tests for each backend on a different file. This way, when testing locally, you can just comment the files for backends you don't want to test for. I can make this modification.

I should have specified that this error only happened in that branch. Whatever the setup is on `master`, `pytest` runs fine even without having `ccl` installed. It must be linked to the modifications made for the OO layer around CT...

m-aguena:
> > @vitenti @combet one way of working around the failures of not having all backends installed it to put the tests for each backend on a different file. This way, when testing locally, you can just comment the files for backends you don't want to test for. I can make this modification.
> 
> I should have specified that this error only happened in that branch. Whatever the setup is on `master`, `pytest` runs fine even without having `ccl` installed. It must be linked to the modifications made for the OO layer around CT...

yes, I made some modifications to test for all combinations of cosmology+modeling backends. That is why it fails if you are missing any backend installed.

clmm/modbackend/func_layer.py
pennalima: Actually, analytic NFW!!!
Numerical NFW also "fails".

combet: So should we explicitly prevent the user from trying those keywords with CCL, or simply add a very loud warning?

pennalima: I think this will be fixed in the next CCL release. So we can include a huge warning: if CCL v?.? use only analytic NFW.

vitenti:
> > > @vitenti @combet one way of working around the failures of not having all backends installed it to put the tests for each backend on a different file. This way, when testing locally, you can just comment the files for backends you don't want to test for. I can make this modification.
> > 
> > 
> > I should have specified that this error only happened in that branch. Whatever the setup is on `master`, `pytest` runs fine even without having `ccl` installed. It must be linked to the modifications made for the OO layer around CT...
> 
> yes, I made some modifications to test for all combinations of cosmology+modeling backends. That is why it fails if you are missing any backend installed.

The tests considering different backends were already happening, all test functions that depend on `modeling_data` parameter will automatically be called once for each backend. You didn't need to load (and shouldn't) any backend directly, the idea is to be transparent to the final user and the tests must check that. That's why I created the pytest fixture that reloads CLMM for each possible backend and runs all tests for each one. If a particular backend is not present the fixture detects it and skips those tests. 

If during the development of a branch you add some `import pyccl`, that will not fail in your machine because you (probably) have CCL installed and pytest will run OK. However, it will break CLMM in any place where CCL is not installed. 

You can try to catch these bugs. I tried a solution where I mess with the import system and make python do not see that CCL is installed (when it actually is), that way the fixture would skip the tests for the CCL backend and test the others (then this is repeated for every backend) and any dangling `import pyccl` would raise an error. The problem with this solution is that the tweaking in the import system does not work in every python version and so I abandoned this solution.

I simpler solution would be to add to the CI different configurations where we install only some backends and run pytest. I have to think in a good way of doing that. We have three backends and therefore 7 possible settings (ct), (nc), (ccl), (ct, nc), (ct, ccl), (nc, ccl), (nc, ct, ccl), I need to find an automated way to iterate over all combinations.

combet:
Thanks for the clarification @vitenti. I agree that all should be automated and transparent to the user. 

@m-aguena, I tried your latest push and once I commented out the `ccl` line in `test_modbackends_oo.py`, the tests ran fine. However, this is less ideal than the previous setup where the tests automatically ran on whatever backend was available.

Suggestion: for v1.0, keep all backends independent from one another and don't perform cosmology objects conversions?
- CT runs with astropy cosmology
- NC with NC cosmo
- CCL with CCL cosmo

Michel, I understand you wanted to allow the use of NC (or CCL?) cosmo with the CT backend, but if the users have a NC cosmo object, they probably have NC installed and should use the NC backend, right?

m-aguena:
> > > > @vitenti @combet one way of working around the failures of not having all backends installed it to put the tests for each backend on a different file. This way, when testing locally, you can just comment the files for backends you don't want to test for. I can make this modification.
> > > 
> > > 
> > > I should have specified that this error only happened in that branch. Whatever the setup is on `master`, `pytest` runs fine even without having `ccl` installed. It must be linked to the modifications made for the OO layer around CT...
> > 
> > 
> > yes, I made some modifications to test for all combinations of cosmology+modeling backends. That is why it fails if you are missing any backend installed.
> 
> The tests considering different backends were already happening, all test functions that depend on `modeling_data` parameter will automatically be called once for each backend. You didn't need to load (and shouldn't) any backend directly, the idea is to be transparent to the final user and the tests must check that. That's why I created the pytest fixture that reloads CLMM for each possible backend and runs all tests for each one. If a particular backend is not present the fixture detects it and skips those tests.
> 
> If during the development of a branch you add some `import pyccl`, that will not fail in your machine because you (probably) have CCL installed and pytest will run OK. However, it will break CLMM in any place where CCL is not installed.
> 
> You can try to catch these bugs. I tried a solution where I mess with the import system and make python do not see that CCL is installed (when it actually is), that way the fixture would skip the tests for the CCL backend and test the others (then this is repeated for every backend) and any dangling `import pyccl` would raise an error. The problem with this solution is that the tweaking in the import system does not work in every python version and so I abandoned this solution.
> 
> A simpler solution would be to add to the CI different configurations where we install only some backends and run pytest. I have to think in a good way of doing that. We have three backends and therefore 7 possible settings (ct), (nc), (ccl), (ct, nc), (ct, ccl), (nc, ccl), (nc, ct, ccl), I need to find an automated way to iterate over all combinations in GHA.

@vitenti I had not realized this dependence on `modeling_data` was implemented, so I guess we are testing for all backends installed and I guess that should be good enough for now. I will change `test_modbackends_oo.py` to work on the same mode.

m-aguena:
> Thanks for the clarification @vitenti. I agree that all should be automated and transparent to the user.
> 
> @m-aguena, I tried your latest push and once I commented out the `ccl` line in `test_modbackends_oo.py`, the tests ran fine. However, this is less ideal than the previous setup where the tests automatically ran on whatever backend was available.
> 
> Suggestion: for v1.0, keep all backends independent from one another and don't perform cosmology objects conversions?
> 
> * CT runs with astropy cosmology
> * NC with NC cosmo
> * CCL with CCL cosmo
> 
> Michel, I understand you wanted to allow the use of NC (or CCL?) cosmo with the CT backend, but if the users have a NC cosmo object, they probably have NC installed and should use the NC backend, right?

@combet I agree, let's keep it simple for v1.0. Each modeling will only accept its predefined cosmology type. We can leave the decision to have the option of mixing cosmologies and modelings for a future discussion. I will opened issue #338 to revisit this later.

combet:
> > Thanks for the clarification @vitenti. I agree that all should be automated and transparent to the user.
> > @m-aguena, I tried your latest push and once I commented out the `ccl` line in `test_modbackends_oo.py`, the tests ran fine. However, this is less ideal than the previous setup where the tests automatically ran on whatever backend was available.
> > Suggestion: for v1.0, keep all backends independent from one another and don't perform cosmology objects conversions?
> > 
> > * CT runs with astropy cosmology
> > * NC with NC cosmo
> > * CCL with CCL cosmo
> > 
> > Michel, I understand you wanted to allow the use of NC (or CCL?) cosmo with the CT backend, but if the users have a NC cosmo object, they probably have NC installed and should use the NC backend, right?
> 
> @combet I agree, let's keep it simple for v1.0. Each modeling will only accept its predefined cosmology type. We can leave the decision to have the option of mixing cosmologies and modelings for a future discussion. I will opened issue #338 to revisit this later.

OK, great - thanks @m-aguena. Let us know when this is ready to be reviewed again!

m-aguena:
@combet @vitenti I made the changes, and the PR is ready for review again.

combet (changes requested):
Thanks @m-aguena! Everything looks good and behaves as expected. There's one last change I'd like to ask for. The "_oo" notebooks (demo_modeling and Example2) specify at the beginning that they will only run with NC or CCL. That's not the case anymore and that introduction text needs to be changed so that CT is also a OO option now.

clmm/modbackend/func_layer.py
m-aguena: @combet @pennalima @vitenti 
I added this check. I set CCL benchmark version to 10 while we wait for a working one.

combet (approved):
All looks good. Thanks @m-aguena.

PR #343, opened 2020-12-04
Issue/341/fix replacement bug

combet:
In `generate_mock_data`, the replacement of "bad" galaxies is failing when there's only one galaxy to replace. In that case, the shape of the column to add into `galaxy_catalog` does no correspond to the shape of the existing column in `galaxy_catalog`. A quick fix was to draw `nbad + 1` galaxies and use only `nbad` of them for the replacement. Thanks @m-aguena for the suggestion :)

### DIFF
--- a/examples/support/mock_data.py
+++ b/examples/support/mock_data.py
@@ -7,6 +7,7 @@ from astropy import units
 from clmm.modeling import predict_tangential_shear, predict_convergence
 from clmm.utils import convert_units, compute_lensed_ellipticity
 
+import pdb
 def generate_galaxy_catalog(cluster_m, cluster_z, cluster_c, cosmo, zsrc, Delta_SO=200, massdef='mean',halo_profile_model='nfw', zsrc_min=None,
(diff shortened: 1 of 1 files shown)


m-aguena (commented):
elegant implementation :+1: 
Did you find the problem with that `seed>100` cluster?

combet:
> elegant implementation 👍
> Did you find the problem with that `seed>100` cluster?

No, haven't looked at it since Friday and I won't have time today or tomorrow. If you want to see the error, try seed=314 in Example3.

m-aguena:
@combet I found the origin of the error: The problem was that the `pzpdf`s usually have different sizes, so they are registered objects with shape `(1,)` on `GCData`. However, for seed=314, the GCData created to replace bad galaxies had only two lines, which coincidentally had the same size pzpdfs (=55), so the shape assigned to each of them was (1, 55).

I managed do fix it, it was not a very elegant solution but it got the job done.

m-aguena:
A better approach was implemented in #347 by @hsinfan1996. I will close this PR and we can solve it there.

PR #344, opened 2020-12-04
Issue/342/separate theory data cosmology

m-aguena:
Created directories to make clear distinctions between data operations and theoretical predictions.
The following directories were created and corresponding files moved to them:

- dataops
- cosmology
- theory
- plotting


combet (approved):
Thanks @m-aguena. I think the code is much more organized this way and this will definitely help both users and the writing of the paper. I checked all the notebooks and ran all the tests without problems. I'm approving this now, so we can then move on to the other issues but we need to remember to address the name of the `theory` folder at some point, as folks were finding it confusing/misleading.

I'll wait until Wed 16/12 5pm (Europe) before merging the PR, to give a chance to @pennalima and/or @vitenti to provide some feedback.

m-aguena:
> Thanks @m-aguena. I think the code is much more organized this way and this will definitely help both users and the writing of the paper. I checked all the notebooks and ran all the tests without problems. I'm approving this now, so we can then move on to the other issues but we need to remember to address the name of the `theory` folder at some point, as folks were finding it confusinNote: Naming of modules will be sorted on #348g/misleading.
> 
> I'll wait until Wed 16/12 5pm (Europe) before merging the PR, to give a chance to @pennalima and/or @vitenti to provide some feedback.

Note: Naming of modules will be sorted on #348

vitenti (approved):
The only modification apart from the renaming was the change in the loading model for the backend, I used relative paths and @m-aguena changed to absolute. The only advantage in using relative paths is that the module can be incorporated inside another python module more easily. But since this is not something that we intend to do I think it's safe to use absolute paths.

combet:
OK great, thanks for the feedback @vitenti. I'll merge it now and we can always revisit absolute versus relative paths if the need arise at some point.

Issue #341, opened 2020-12-03
mock_data.py: generate_galaxy_catalog() error


hsinfan1996:
#335 
**This occurs in issue/335/nb_for_paper1.0/examples/Paper_v1.0.ipynb,** which is going to be used in the paper. It hints that there may be a bug although old examples do not have the issue (at least for now).
When photoz_sigma_unscaled is turned on and there is only one bad galaxy in the generated galaxy catalog, Astropy returns ValueError: shape mismatch. The problem is down to line 144 galaxy_catalog[badids] = replacements. When there is only one galaxy to be replaced, somehow assigning the 'pzbins' and 'pzpdf' columns will fail and results in the error.
To reproduce, set the random seed to one of the following in issue/335/nb_for_paper1.0/examples/Paper_v1.0.ipynb
random seeds with 1 bad galaxy: 13, 35, 40, 54, 60, 68, 82, 84, 92, 94

No error reported for the following:
random seed(s) with > 1 bad galaxies: 41

hsinfan1996:
Example3 is also affected, with different random seeds. Seed 17 will cause the problem.

hsinfan1996:
It is really hard to understand why seed 341 did not work after the fix, because it seems to be the same problem as other seeds. 
If the data of the replacements are copied one by one to replace the bad galaxies in the galaxy catalog, there is no error reported from seed 0 to 1000. But it would be more like an ad-hoc fix.
```
replacements = _generate_galaxy_catalog(ngals=nbad, **params)
for column in galaxy_catalog.colnames:
    for n, badid in enumerate(badids):
        galaxy_catalog[column][badid]=replacements[column][n]
```

PR #347, opened 2020-12-10
issue/341/fix replacement error: a different approach

hsinfan1996:
Copy the fields one by one to prevent broadcasting, which causes the error.


m-aguena:
This is a much better approach to it. I will close PR #343 and issue #341 will be solved here.

m-aguena (approved):

combet:
Thanks @hsinfan1996  and @m-aguena. This looks good; I'll merge it

hsinfan1996:
@m-aguena @combet Thank you

Issue #325, opened 2020-11-20
Names of modeling functions in the functional and OO interface


combet:
Functions to predict lensing quantities do not have the same names when using the functional or the object-oriented interface. Need to decide on a consistent set of names to be implemented for v1.0.

m-aguena:
FYI, those are the current names
```
Functional                        |  Object Oriented
---------------------------------------------------------
get_3d_density                    |  eval_density
predict_surface_density           |  eval_sigma
predict_excess_surface_density    |  eval_sigma_excess
get_critical_surface_density      |  eval_sigma_crit
predict_tangential_shear          |  eval_shear
predict_convergence               |  eval_convergence
predict_reduced_tangential_shear  |  eval_reduced_shear
predict_magnification             |  eval_magnification
```

combet:
Thanks for the summary @m-aguena. I like the more explicit name of the functional interface (useful for people new to the field) and the more consistent and compact naming convention of the OO interface. So, as a first suggestion and to kick-start the discussion, I'll simply suggest a mixture of the two:
```
eval_3d_density
eval_surface_density
eval_excess_surface_density
eval_critical_surface_density
eval_tangential_shear
eval_convergence
eval_reduced_tangential_shear
eval_magnification
```

vitenti:
I was thinking about the naming convention. I normally use eval (evaluate) to mean: given the object representing a model (here model can range from an actual model to a simple integrand) we can evaluate different aspects of it. So in short, eval means that the context is given by the object and you are going to just evaluate some quantity. In my opinion, "predict" is a little bit off, usually, it implies some kind of prediction given some data and that's not what it does. I would use "compute", meaning: given the input compute the value of something. 

I suggest the following:

- eval: used in methods to imply that it is an evaluation in the context of the object.
- compute: used in the functional interface informing that the computation depends only on the input parameter.

About the names themselves, we have two options, in the first (that we adopt in NumCosmo), we have in the documentation of each object an introduction where all quantities are discussed and defined, then the methods carry the names of the symbols used in this introduction. The second option is to avoid this split and put all documentation in each method, in this case, the method names should be more informative (as in the suggestion @combet gave above). I think that python module documentation usually follows the second approach. All in all, we should add a section to documentation explaining all these choices.

combet:
[...]

PR #352, opened 2020-12-17
Issue/325/rename modeling functions

combet:
Homogenise the function/method names for the OO and functional interfaces of the theory module. Use decision made in #325.

### DIFF
--- a/clmm/__init__.py
+++ b/clmm/__init__.py
@@ -3,7 +3,7 @@ from .gcdata import GCData
 from .galaxycluster import GalaxyCluster
 from .dataops import compute_tangential_and_cross_components, make_radial_profile
 from .utils import compute_radial_averages, make_bins, convert_units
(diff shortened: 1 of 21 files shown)


m-aguena (changes requested):
It all seems good, according to the plan. I just think it is missing the update on the pair `get_critical_surface_density, eval_sigma_crit`

combet:
> It all seems good, according to the plan. I just think it is missing the update on the pair `get_critical_surface_density, eval_sigma_crit`

Done.

m-aguena (changes requested):
There is only one more to go, this one we even missed listing on the issue:
`get_reduced_shear_from_convergence` in `clmm/theory/generic.py`

combet:
> There is only one more to go, this one we even missed listing on the issue:
> `get_reduced_shear_from_convergence` in `clmm/theory/generic.py`

And done :)

m-aguena (approved):
Great! Thanks

m-aguena:
I will merge this PR to move things along, and if we can open another isssue later if we still find the naming of something can be improved.

PR #354, opened 2020-12-17
Issue/307/nan to num

m-aguena:
Remove nan to num, but add safeguard on eval_sigma_critical.

### DIFF
--- a/clmm/theory/parent_class.py
+++ b/clmm/theory/parent_class.py
@@ -1,6 +1,5 @@
 # CLMModeling abstract class
 import numpy as np
-import warnings
 
 
 class CLMModeling:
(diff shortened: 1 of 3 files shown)


combet (approved):
It looks good, thanks @m-aguena!

clmm/theory/parent_class.py
```diff
@@ -122,6 +121,10 @@ def eval_sigma_crit(self, z_len, z_src):
         float
             Cosmology-dependent critical surface density in units of :math:`M_\odot\ Mpc^{-2}`
         """
+        if z_len<=0:
+            raise ValueError(f'Redshift for lens <= 0.')
+        if np.any(np.array(z_src)<=0):
```
pennalima, line 126: Hi @m-aguena  and @combet, 
shouldn't the safeguard be z_src <= z_lens?

combet: Hi @pennalima - we discussed that, but no, we want to keep the possibility to have z_src<z_lens as to generate mock data, it allows to contaminate the sample with foreground galaxies (which does happen in real life). When z_src<z_lens, the predictions safely go zero for shear and convergence, so no problem there.

pennalima (approved):

pennalima:
@combet You can merge. :)

PR #360, opened 2021-01-19
Issue/358/dc2 examples

hsinfan1996:
NOTE: According to https://github.com/LSSTDESC/gcr-catalogs, dc2_object_run2.2i_dr3 is not fully completed.

### DIFF
--- a/examples/DC2_examples/DC2_gt_profiles.ipynb
+++ b/examples/DC2_examples/DC2_gt_profiles.ipynb
@@ -4,7 +4,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# Compare tangential shear profiles from the extragalactic and object catalogs for DC2 Run 2.1i\n",
+    "# Compare tangential shear profiles from the extragalactic and object catalogs for DC2 Run 2.2i\n",
     "\n",
     "This notebook can be run at NERSC or CC-IN2P3 where the DESC DC2 products are stored. You need to be a DESC member to be able to access those. The DC2 catalog-related imports below (`FoFCatalogMatching`, `GCR` and `GCRCatalogs`) are readily available from the `desc` conda environement at NERC or CC-IN2P3. If working outside such environment, these packagea first need to be installed. \n",
     "\n",
@@ -53,7 +53,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "object_cat = GCRCatalogs.load_catalog('dc2_object_run2.1i_dr1')"
+    "object_cat = GCRCatalogs.load_catalog('dc2_object_run2.2i_dr3')"
    ]
   },
   {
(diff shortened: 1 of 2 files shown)


examples/DC2_examples/DC2_gt_profiles.ipynb
```diff
@@ -396,7 +396,7 @@
    "outputs": [],
    "source": [
     "dc2_cosmo = extragalactic_cat.cosmology\n",
-    "cosmo = clmm.Cosmology(H0 = dc2_cosmo.H0, Omega_dm0 = dc2_cosmo.Om0-dc2_cosmo.Ob0, Omega_b0 = dc2_cosmo.Ob0)\n",
+    "cosmo = clmm.Cosmology(H0 = dc2_cosmo.H0.value, Omega_dm0 = dc2_cosmo.Om0-dc2_cosmo.Ob0, Omega_b0 = dc2_cosmo.Ob0)\n",
```
combet, line 399: @hsinfan1996 - I just ran the notebook from master, i.e. without the change you made to `dc2_cosmo.H0` and it ran fine. So, I'm a bit puzzled as to why that change is needed... You ran at NERSC?

hsinfan1996: I ran at NERSC and on my laptop, and they all gave errors. I am thinking if this problem is unique to me too.

examples/DC2_examples/DC2_gt_profiles.ipynb
```diff
@@ -53,7 +53,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "object_cat = GCRCatalogs.load_catalog('dc2_object_run2.1i_dr1')"
+    "object_cat = GCRCatalogs.load_catalog('dc2_object_run2.2i_dr3')"
```
combet, line 56: Actually, let's make the default `dc2_object_run2.2i_dr6`. I corresponds to 5 years of LSST data (while dr3 is 2 years) and there doesn't seem to be any specific warning with that catalog.

examples/DC2_examples/DC2_gt_profiles.ipynb
shenmingfu: I tested it on NERSC and I found I had to use `dc2_cosmo.H0.value`, because otherwise it would give an error related to ccl.
I used `desc-python` rather than `conda-clmmenv` to import gcr catalogs, and I adjusted CLMM imports by adding 
```
import sys
sys.path.append('../../')
```

combet: Thanks @shenmingfu. So I finally managed to reproduced the error and it happens with the CCL and NC backends. The cluster-toolkit backend is fine because it uses Astropy cosmology object which can handle `quantities`, but not the two other backends. So yes, we need `dc2_cosmo.H0.value` for it to work with all cosmo backends.

Thanks for spotting the issue @hsinfan1996!

hsinfan1996: oh great, thank you. @combet @shenmingfu

combet (approved):
Thanks @hsinfan1996, all looks good. Approving PR :)

PR #363, opened 2021-01-20
Issue/361/nb plot display

m-aguena:
Fix use of matplotlib `Agg` to testing only.


combet (approved):
Thanks @m-aguena. Display is working fine again on jupyter plateforms.

Issue #369, opened 2021-01-28
Make massdef and halo_profile_model ignore letter casing


m-aguena:
Make `halo_profile_model` work with both `nfw` and `NFW`, and other configs. Same for `massdef`.

PR #370, opened 2021-01-28
add lowercase fix

m-aguena:
From Issue #369


combet (commented):
Thanks @m-aguena - worked like a charm.

Question: should we update the docstrings to reflect the fact that upper case are now permitted?

m-aguena:
@combet that is a good idea

m-aguena:
@combet I removed the case dependence of `error_model` in `compute_radial_averages` and `method` in `make_bins` and added it to the documentation.

combet (approved):
All good, thanks @m-aguena!

Issue #366, opened 2021-01-28
Small things to correct/enhance


marina-ricci:
Documentation :
1. - [ ] in clmm.theory.compute_* => "z_src_model" keyword : not clear from the documentation what it is doing.
 
. Enhancement : 
1. - [x] in clmm.theory.compute_* =>  allow to write either "NFW" or "nfw" (#370 )
 3.  - [ ] clmm.GalaxyCluster.make_radial_profile : we should be able to chose the error model : in case the noise is not shape noise dominated the errors shouldn't be computed as std/np.sqrt(N_src)
 
Bugs ?:
1. - [x] in example notebooks : we say the theory module works with "h" where it is not the case (done in #345 - no need for a new branch for this small change)
 5.  - [x] clmm.support.mock data not accessible from clmm.support need to import it explicitly (done in #368 )
 6.  - [x] when accessing e.g. cl.make_radial_profile we can not access the function documentation

combet:
Thanks for this @marina-ricci!
- We have an open PR for the documentation (#365), so the first point can be taken care of in that branch and for v1.0. If you could take care of that'd be great.
- I'd suggest we leave the enhancements as such for the time being. We mention in the paper, below eq 12 that we will improve on this soon (that's clearly a limitation as of now)
- We should definitely fix any mention of 'h' in the notebooks for v1.0. Could you also take care of it?

marina-ricci:
OK. 
What about points 2 and 3 of "Bugs".
`clmm/support/__init__.py` is empty. I guess this causes point 2.

m-aguena:
I created a quick fix for `clmm.support.mock data not accessible from clmm.support need to import it explicitly` - PR #368 

@marina-ricci I do not see this error `when accessing e.g. cl.make_radial_profile we can not access the function documentation`. How are you getting that?

marina-ricci:
Thank's @m-aguena!
- I just checked and (3) is quite subtle. 

> 

> `test = [clmm.GalaxyCluster('mock', cluster_ra, cluster_dec, 
                               cluster_z, mock_catalogs[i]), clmm.GalaxyCluster('mock', cluster_ra, cluster_dec, 
                               cluster_z, mock_catalogs[i])]

test1 = clmm.GalaxyCluster('mock', cluster_ra, cluster_dec, 
                               cluster_z, mock_catalogs[i])`

m-aguena:
@marina-ricci I am not sure the reason, but this syntax does not work.
[...]

PR #373, opened 2021-02-12
Issue 366

marina-ricci:
This PR addresses the remaining points in issue #366 :

1. - I removed mention to "h" difference between data and theory in the Example notebooks.
2. - I modified the description of the `z_src_model ` keyword in the docstrings of `func_layer.py` and I added that `r_proj` can also be a float.

Regarding (2) : I removed the mention of the non-yet supported options for  `z_src_model` and I kept them as a TODO note.


combet (approved):
Thanks @marina-ricci! All changes look good.

PR #383, opened 2021-03-23
Issue/377/srd z distrib

combet:
Add the DESC SRD redshift distribution to mock data, following suggestion from internal review of v1.0 paper.
- [x] New function and corresponding options and docstring  added to `clmm/support/mock_data.py`
- [x] Corresponding test included in `tests/test_mock_data.py`

Note: I first replicated the test from Chang et al, which did not work at all with the new parameters. Then realized I did not understand how the test for Chang et al. was designed, so updated it to a more generic form.


m-aguena (approved):
Looks good to me! (we should add some missing documentation to this mock_data module in the future though)

shenmingfu (approved):
They look great to me. Thanks.

PR #388, opened 2021-04-02
Add computations without flat sky

m-aguena:
Computations for (theta, phi) without flatsky approximations. The notebook (Flat_sky_limitations.ipynb) shows a difference at 0.0001% level on the mass fitted (10,000 times smaller than the mass error).

### DIFF
--- a/clmm/dataops/__init__.py
+++ b/clmm/dataops/__init__.py
@@ -110,7 +110,7 @@ def compute_tangential_and_cross_components(
         angsep, phi = _compute_lensing_angles_flatsky(ra_lens, dec_lens,
                                                       ra_source_, dec_source_)
     elif geometry=='curve':
-        angsep, phi = _compute_lensing_angles_ap(ra_lens, dec_lens,
+        angsep, phi = _compute_lensing_angles_astropy(ra_lens, dec_lens,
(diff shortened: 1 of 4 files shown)


m-aguena:
The tests are crashing because we computed the values assuming flat sky approximations. We should either recompute the values in the tests, increase the allowed error or pass flat sky to the tests.

combet:
Thanks @m-aguena - this looks good.

> The tests are crashing because we computed the values assuming flat sky approximations. We should either recompute the values in the tests, increase the allowed error or pass flat sky to the tests.

Given that the default is now `curve`, I'd be in favour of having that option tested at least minimally. I've tried increasing the tolerance by a factor ten, but it's still not enough for the tests to pass. So maybe a compromise would be to:
- use the `flat` option when the tests call `compute_tangential_and_cross_components`
- add a `test_compute_lensing_angles_ap` that tests the curve option and maybe compares to the flatsky results (with a loose tolerance, 1e-2 might be sufficient)

What do you think?

m-aguena:
@combet I fixed the tests to use flat_sky and created issue #390 to implement tests of curve sky in the future.

combet (approved):
Thank you @m-aguena, all good!

PR #393, opened 2021-04-30
add checks for (r/m/c)<=0 in all profile functions

m-aguena:


combet (approved):
Thanks @m-aguena, all looks good,

Issue #403, opened 2021-06-12
Error messages in mock_data


cristobal-sifon:
For a bunch of arguments in `mock_data.generate_galaxy_catalog` (most of them, actually) we do not assess their values and they end up producing errors midway through the code, with messages that can be quite cryptic for the user. This is probably most important with reference to For instance,

```python
mock.generate_galaxy_catalog('1', cluster_z, concentration, cosmo, src_z, ngals=10000)
```
produces
```python
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-17-97fe3aafab64> in <module>
----> 1 mock.generate_galaxy_catalog('1', cluster_z, concentration, cosmo, src_z, ngals=10000)

~/anaconda3/envs/clmm/lib/python3.8/site-packages/clmm/support/mock_data.py in generate_galaxy_catalog(cluster_m, cluster_z, cluster_c, cosmo, zsrc, Delta_SO, massdef, halo_profile_model, zsrc_min, zsrc_max, field_size, shapenoise, mean_e_err, photoz_sigma_unscaled, nretry, ngals, ngal_density)
    147         ngals = _compute_ngals(ngal_density, field_size, cosmo, cluster_z, zsrc, zsrc_min=zsrc_min, zsrc_max=zsrc_max)
    148 
--> 149     galaxy_catalog = _generate_galaxy_catalog(ngals=ngals, **params)
    150     # Check for bad galaxies and replace them
    151     nbad, badids = _find_aphysical_galaxies(galaxy_catalog, zsrc_min)

~/anaconda3/envs/clmm/lib/python3.8/site-packages/clmm/support/mock_data.py in _generate_galaxy_catalog(cluster_m, cluster_z, cluster_c, cosmo, ngals, zsrc, Delta_SO, massdef, halo_profile_model, zsrc_min, zsrc_max, shapenoise, mean_e_err, photoz_sigma_unscaled, field_size)
    246     galaxy_catalog = _draw_galaxy_positions(galaxy_catalog, ngals, cluster_z, cosmo, field_size)
    247     # Compute the shear on each source galaxy
--> 248     gamt = compute_tangential_shear(galaxy_catalog['r_mpc'], mdelta=cluster_m,
    249                                     cdelta=cluster_c, z_cluster=cluster_z,
    250                                     z_source=galaxy_catalog['ztrue'],

[...]

Issue #394, opened 2021-05-25
Add measurement errors in mock data


combet:
At the moment, mock data does not allow to generate measurement errors on the observed ellipticities, contrarily to what shape measurements codes produce. This should be implemented.

cristobal-sifon:
If we add this we should also probably add functionality to deal with it in ``clmm``. What would be the plan for this? And also, I worry that we would need to consider correlation in uncertaintes between redshifts and ellipticities, unless we make it very clear that the resulting catalog cannot be subsampled in any way (save randomly?)

combet:
@cristobal-sifon, yes of course of need to deal with it in `clmm`. So in my mind, the first place where those errors would be used would be in weights to be added to the evaluation of the shear profile (g_t or DeltaSigma), which is somewhat contained in #395 for which @m-aguena already opened a PR. It is also linked to #237 (which was open a while ago). Providing the actual weights for the DeltaSigma profile, which use the error on the shape but not only (see eqs in #396), has not yet been flag as an issue, but should be (feel free to open an new issue). Am I forgetting other places where these errors would come into play?

Regarding your second point, could you clarify please? do you mean that there is a correlation as more distant galaxies will generally have large ellipticity uncertainties?

cristobal-sifon:
Thanks @combet 

> Regarding your second point, could you clarify please? do you mean that there is a correlation as more distant galaxies will generally have large ellipticity uncertainties?

yes that's what I mean, broadly speaking. More specifically both the redshift and its uncertainty, as well as the uncertainty on the ellipticity, depend on the brightness and size of the galaxy - and survey depth, of course. This is not a simple issue, and we should consider leaving it out of `clmm`. However if we really want to have this I would suggest reaching out to HSC to see if we can get a realistic joint distribution of all these quantities (not the actual catalog but an n-dimensional histogram) from which to draw them for our mock samples. Would be good to discuss

combet:
Thanks for the clarification @cristobal-sifon. I agree that this is a complex problem. My initial thought was to generate shape errors independently of redshifts in `mock_data`. That would not be realistic, but would allow us to exercise the other items/steps of `clmm` that do need to account for those errors. In parallel, we could indeed try to get the information we'd need to make this realistic, but this could be seen as a second step. Let's discuss this is more details at a future tag-up/co-work session.

Also pinging @shenmingfu who first suggested that addition to `mock_data`.

cristobal-sifon:
[...]

PR #406, opened 2021-06-13
addresses #394 and #403

cristobal-sifon:
Adds the posibility of having ellipcitity uncertainties when generating mock data, and provides more useful error messages when calling `mock_data.generate_galaxy_catalog` with the wrong arguments.

However, I had some trouble compiling the updated docs, and I'd like opinions on whether we want to include this in the mass fitting demos.

### DIFF
--- a/clmm/support/mock_data.py
+++ b/clmm/support/mock_data.py
@@ -3,14 +3,20 @@ import numpy as np
 import warnings
 from scipy import integrate
 from scipy.interpolate import interp1d
+from scipy.stats import chi2
 from astropy import units
 
 from clmm import GCData
 from clmm.theory import compute_tangential_shear, compute_convergence
 from clmm.utils import convert_units, compute_lensed_ellipticity
 
-def generate_galaxy_catalog(cluster_m, cluster_z, cluster_c, cosmo, zsrc, Delta_SO=200, massdef='mean',halo_profile_model='nfw', zsrc_min=None,
-                            zsrc_max=7., field_size=8., shapenoise=None, photoz_sigma_unscaled=None, nretry=5, ngals=None, ngal_density=None):
+
+def generate_galaxy_catalog(cluster_m, cluster_z, cluster_c, cosmo, zsrc,
+                            Delta_SO=200, massdef='mean',
+                            halo_profile_model='nfw', zsrc_min=None,
+                            zsrc_max=7., field_size=8., shapenoise=None,
+                            mean_e_err=None, photoz_sigma_unscaled=None,
+                            nretry=5, ngals=None, ngal_density=None):
     r"""Generates a mock dataset of sheared background galaxies.
 
     We build galaxy catalogs following a series of steps.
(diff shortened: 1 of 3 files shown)


clmm/support/mock_data.py
```diff
@@ -95,6 +101,11 @@ def generate_galaxy_catalog(cluster_m, cluster_z, cluster_c, cosmo, zsrc, Delta_
         Proper distance in Mpc  at the cluster redshift.
     shapenoise : float, optional
         If set, applies Gaussian shape noise to the galaxy shapes with a width set by `shapenoise`
+    mean_e_err : float, optional
+        Mean per-component ellipticity uncertainty. Currently,
+        individual uncertainties are drawn from a uniform distribution
+        in the range [0.9,1.1]*mean_e_err. If not provided, the output
```
m-aguena, line 107: Is there a reason to have this specific range here? I mean is setting this choice of 10% variation arbitrary?

m-aguena:
I like the idea of testing the types of the aguments in our functions. We could think about adding this test to other functions of clmm. @combet @aimalz what do you think?
We can generalize  the tests in `_test_input_params` and apply to more api functions.

clmm/support/mock_data.py
cristobal-sifon: totally arbitrary. We decided we wanted any implementation to get us going, untiil we have something even remotely realistic. Happy to update to your preferred range/distribution

combet:
Thanks @cristobal-sifon! For this particular issue/PR, I think updating the `demo_generate_mock_cluster` to show this new functionality might be sufficient. It needs to be added to the mass fitting examples, but I think this could be done in a new issue, where we'll use both this functionality and the weighted average one from #399. Do you agree?

@m-aguena, I agree that generalizing checking the argument types would be a nice improvement.

cristobal-sifon:
I apologize about this new commit addressing a different issue, but by mistake I branched off of the branch of this PR instead of `master` and I didn't know how to revert that (and there's stuff that was changed originally upon which this last commit depended). It's hopefully a small enough commit that it doesn't cause too much trouble

cristobal-sifon:
> Thanks @cristobal-sifon! For this particular issue/PR, I think updating the `demo_generate_mock_cluster` to show this new functionality might be sufficient. It needs to be added to the mass fitting examples, but I think this could be done in a new issue, where we'll use both this functionality and the weighted average one from #399. Do you agree?

That sounds good, in which case I will mark this PR as ready for review (please revert if that's not actually the case)

clmm/support/mock_data.py
m-aguena: This is fine for now, we can re-address this when implementing more realistic error in the future.

m-aguena:
@cristobal-sifon I removed your changes regarding the PDF from this branch to and moved them to [issue/404/galcat_pdf_improvement](https://github.com/LSSTDESC/CLMM/tree/issue/404/galcat_pdf_improvement)

clmm/support/mock_data.py
```diff
@@ -3,14 +3,20 @@
 import warnings
 from scipy import integrate
 from scipy.interpolate import interp1d
+from scipy.stats import chi2
```
hsinfan1996, line 6: Is there any specific reason for importing this function? I did not find the use of this function in the rest of the file.

cristobal-sifon: you are right. I was testing ways to implement error distributions and forgot to remove it. Thanks for catching that!

PR #409, opened 2021-06-24
improve print and display of GCData and GalaxyCluster

m-aguena:


m-aguena:
@cristobal-sifon what do you think of this version of `print(GalaxyCluster)`?

cristobal-sifon:
looks good! I would change one tiny detail, though this is personal taste only.

```python
#print(ideal_data)
GCData
> defined by:cosmo=None
> with columns: ra, dec, e1, e2, z, ztrue, id
> and 1000 objects
          ra                   dec                    e1           ... ztrue  id
--------------------- --------------------- ---------------------- ... ----- ---
  0.18131552891208902 -0.051521602053583054  -0.009481107480924065 ...   0.8   0
# ...
```
while
```python
#print(gc_object)
GalaxyCluster Awesome_cluster: (ra=0.0, dec=0.0) at z=0.3
> 10000 source galaxies
> With columns: ra dec e1 e2 z ztrue id

          ra                   dec          ... ztrue  id 
--------------------- --------------------- ... ----- ----
  0.15851176371796402  -0.15151992029762024 ...   0.8    0
# ...
```
so I would print both consistently, probably first the columns and then the number of objects/source galaxies, with "with columns" all lower case in both cases, and possibly the cosmology as well. Oh, and something even more pedantic, while I'm on it: there's a space missing after the colon after "defined by".

It might be useful to define something like a `_print_tbl` helper called in the `__str__` of both these classes so that this can be unified and extended easily if necessary. Not critical though, of course

cristobal-sifon:
Thanks! everything looks good to me

combet (approved):
Thanks @m-aguena, all looks good

Issue #416, opened 2021-07-12
Import `compute_magnification` missing. 


marina-ricci:
In `clmm/__init__.py` the import of `theory.compute_magnification` is missing.

PR #417, opened 2021-07-12
Issue 416

marina-ricci:
This PR address issue #416

### DIFF
--- a/clmm/__init__.py
+++ b/clmm/__init__.py
@@ -3,7 +3,7 @@ from .gcdata import GCData
 from .galaxycluster import GalaxyCluster
 from .dataops import compute_tangential_and_cross_components, make_radial_profile
 from .utils import compute_radial_averages, make_bins, convert_units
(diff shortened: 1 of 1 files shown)


m-aguena (approved):

combet:
@m-aguena - looking at the single file that has changed in this PR, I noticed that the version number is listed as both 0.9.3 in the "old" and "new" version of the file, while master is currently at 0.9.4. While this is super minor in principle (just a matter of merging master again in that branch?) I just want to double check with you that it doesn't mean something else went unseen when you fixed the "bad commit" into master the other day. Sorry for being extra-cautious (i already broke master once, don't want to do it again...). Please give me the green light and I'll go ahead and merge.

m-aguena:
@combet that is ok. This branch was created before v0.9.4, so it is expected to have a previous versioning on it. As long as git does not think there was any change in the specific line that has the version, it should merge it automatically. Thanks for checking though!

PR #435, opened 2021-09-10
Issue/433/credits readme

cavestruz:
Added credits, acknowledgments, readme; closing Issue 433.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,3 +72,82 @@ Contributing documentation can be found [here](CONTRIBUTING.md)
 
 ## Contact <a name="contact"></a>
 If you have comments, questions, or feedback, please [write us an issue](https://github.com/LSSTDESC/CLMM/issues).
+
+
+# License, Credits, Feedback etc
+
+This code has been released by DESC, although it is still under active
+development.  It is accompanied by a journal paper that describes the
+development and validation of `CLMM`, which you can find on the
+arxiv:[2107.10857](https://arxiv.org/abs/2107.10857).  If you make use
+of the ideas or software here, please cite that paper and provide a
+link to this repository: https://github.com/LSSTDESC/CLMM. You are
+welcome to re-use the code, which is open source and available under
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -72,3 +72,82 @@ Contributing documentation can be found [here](CONTRIBUTING.md)
...
+link to this repository: https://github.com/LSSTDESC/CLMM. You are
+welcome to re-use the code, which is open source and available under
+terms consistent with our
+[LICENSE](https://github.com/LSSTDESC/CLMM/blob/main/LICENSE) ([BSD
+3-Clause](https://opensource.org/licenses/BSD-3-Clause)).
+
+External contributors and DESC members wishing to use CLMM for
+non-DESC projects should consult with the CL working group conveners,
+ideally before the work has started, but definitely before any
+publication or posting of the work to the arXiv.
```
combet, line 93: I see there is something similar in the CCL repo, but this doesn't really make sense unless `non-DESC project` becomes `DESC project`. We probably need to go back to the software policy to double check things.

combet (changes requested):
Thanks @cavestruz - there's just one thing that needs to be double-checked, I think.

README.md
cavestruz: Ah, you're right.  I've changed the text accordingly (it absolutely makes much more sense for it to be `DESC project`).  But, it occurred to me that if folks in the DESC community want to use CLMM in general, it would be nice to let the co-conveners know they are using the DESC product.

cavestruz: Notes from tag-up:

- Would be good for folks in DESC community looking to use CLMM in a non-DESC project, it would be good practice to contact CL co-conveners and the CLMM team leads to let them know of your plans.  A desired outcome of this is that your non-DESC project concept using CLMM can be presented to the CL WG to identify things that might mutually benefit the non-DESC project and ongoing DESC projects since both involve folks in the community.

PR #440, opened 2021-10-05
Issue/439/use comp red shear from generic

m-aguena:
Using `theory.generic.compute_reduced_shear_from_convergence` function for computations in classes (excluding numcosmo). I also noted there were other functions in `cluster_toolkit` and `ccl` backends that only used internal generic products, so I moved them to the parent class.

### DIFF
--- a/clmm/theory/ccl.py
+++ b/clmm/theory/ccl.py
@@ -143,32 +143,5 @@ class CCLCLMModeling(CLMModeling):
         dens = self.hdpm.projected(*args)
         return (mean_dens-dens)*self.cor_factor/a_cl**2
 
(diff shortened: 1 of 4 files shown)


m-aguena:
@vitenti in @marina-ricci 's implementation of the magnification it is assumed sprerical symmetry (i. e. zero cross shear). Is the same assumed in NumCosmo?

m-aguena:
@marina-ricci I understand what was made in your implementation. I just wanted to know if the same thing is beind done in NumCosmo. @vitenti ?

vitenti:
@m-aguena Yes, currently we implemented the computations for spherically symmetric profiles. In the future we might have to implement non-spherically symmetric profiles too.

PR #441, opened 2021-10-08
Issue/392/boost functionality

cavestruz:


combet:
@chrisdavies234, @cavestruz - Thanks for this new functionality! I just started looking at the new demo notebook from Chirs and I got puzzled by the fact that the boost factor lowered the values of the data vector. I think there's a bit of definition to be careful about regarding whether we multiply of divide by the boost factor. My understanding is that:
- either the boost factor is applied to the data vector to correct for the dilution of the signal by member galaxies, in which case the corrected profile should have a high value
- or the boost factor is applied to the model prediction, in which case it is really a dilution factor, and the resulting model prediction would be lower that the original one.

In this notebook, the data is generated with `mock_data` (without dilution as we do not have that functionality), but the corrected profiles from those data are lower than the uncorrected one.

So maybe the argument should be reverted and say. Here's a mock profile without dilution and by applying this "boost factor", we're diluting it to mimick the effect of contamination by cluster members.  

I don't know if the above is understandable. Might be easier to chat about it at our next co-wok next year. But as it is, the notebook might be a bit confusing I think.

cavestruz:
Ah, so referring to the example from [cluster_toolkit](https://cluster-toolkit.readthedocs.io/en/latest/source/boostfactors.html), it looks like the boost factor is the _divisor_.  So, the uncorrected profile has a larger normalization than any of the boost-corrected profiles.  I think I agree with @combet's phrasing that 

> here's a mock profile without dilution and by applying this "boost factor", we're diluting it to mimick the effect of contamination by cluster members.

I think the specific suggested change here is to modify the phrasing in the <demo_boost_factors.ipynb>?

e.g. 

- [ ] Modify phrasing in <demo_boost_factors.ipynb> to clarify how boost factors are used and that we expect the boost-correction to lower the normalization of the uncorrected profile.  
- [ ] Clarify figure legend labels so it is easy/quick to see the divisor, e.g. [cluster_toolkit example](https://cluster-toolkit.readthedocs.io/en/latest/source/boostfactors.html)

@chrisdavies234 - I'd be happy to hop on and make these changes, unless you'd prefer to do them since you are the original author of the notebook.

chrisdavies234:
I have updated the notebook, which should be clearer now. If the legend labels are not updated, let me know - i was having a bit of trouble with git

m-aguena:

marina-ricci (approved):

marina-ricci:
This looks good to me !

Issue #448, opened 2021-10-25
CLMM on real data


shenmingfu:
This issue is about using CLMM on real datasets. The details are given in https://github.com/LSSTDESC/SprintWeek2021/issues/51 .

combet:
This failed to closed automatically when the PR was merged. Closing it now.

PR #450, opened 2021-10-26
Running CLMM on real datasets

shenmingfu:
This is for issue #448 . The new branch is issue/448/read_data. I made a notebook `examples/read_data.ipynb` and it's still being built. I made this pull request so other people would be able to see it.


marina-ricci:
@shenmingfu thank you for this nice notebook. It runs well on my side, I only have minor comments to improve clarity and one about input data.

As we discussed, it would be great to have a way to download an example catalog from the e.g the nersc lsst webpage, such as the user would only have 1 command to run, and nothing to change from the notebook. 
You can still provide the info to download the data from the HSC portal in this notebook or somewhere else. Following naively your notebook, I wasn't sure were exactly to go on the HSC webpage to download the data. Also the query only works for PR1 and you have to change manually (the default in PR3), but it was not stated explicitly in your notebook. Finally, please also provide the command to unzip the data.

marina-ricci (changes requested):
@shenmingfu thank you for this nice notebook. It runs well on my side, I only have minor comments to improve clarity and one about input data.

As we discussed, it would be great to have a way to download an example catalog from the e.g the nersc lsst webpage, such as the user would only have 1 command to run, and nothing to change from the notebook.
You can still provide the info to download the data from the HSC portal in this notebook or somewhere else. Following naively your notebook, I wasn't sure were exactly to go on the HSC webpage to download the data. Also the query only works for PR1 and you have to change manually (the default in PR3), but it was not stated explicitly in your notebook. Finally, please also provide the command to unzip the data.

marina-ricci:
On the profile construction : 

- [ ] - Please provide reference for the apply_shear_calibration function in the comments.
- [ ] -  Also add a comment for the "photo_z" type cell.
- [ ] Consider applying a cut at z_cl +0.1 (or greater) on the photoz
- [ ] I don't understand the need of the  ra-dec selection you make.
- [ ] I don't understand the need of the "# Shear profile  Using clmm.dataops functions " cell, since you don't use it after 
- [ ]

shenmingfu:
Hi @marina-ricci , the reason for the ra-dec is selection is that the query is made for the whole field and the area cut can simplify the processing. It seems that adding an area cut in the query would slow down the query. 
"clmm.dataops functions" is used in the original notebook for paper v1.0 to show the functionalities of CLMM (functional vs object-oriented programming). Here we can still test it using real data.
I added comments in the notebook for them.

examples/real_data.ipynb
```diff
@@ -0,0 +1,769 @@
...
+    "# Shear profile\n",
+    "# Using/testing clmm.dataops functions (details are in the notebook for Paper V1.0).\n",
+    "# This step is for testing the CLMM functionalities; we will not use the result in later step.\n",
+    "\n",
+    "import clmm.dataops as da\n",
+    "# Convert elipticities into shears.\n",
+    "gal_ang_dist, gal_gt, gal_gx = da.compute_tangential_and_cross_components(cluster_ra, cluster_dec,\n",
+    "                                                                          gal_ra, gal_dec,\n",
+    "                                                                          gal_e1, gal_e2,\n",
+    "                                                                          geometry=\"flat\")\n",
```
combet, line 397: Please, remove `geometry='flat'`. The default is now to use the correct `curve` geometry, using astropy functionality.

examples/real_data.ipynb
```diff
@@ -0,0 +1,769 @@
...
+    "galaxies = clmm.GCData([gal_ra, gal_dec, gal_e1, gal_e2, gal_z, gal_id],\n",
+    "                      names=['ra', 'dec', 'e1', 'e2', 'z', 'id'])\n",
+    "\n",
+    "# Create a GalaxyCluster.\n",
+    "cluster = clmm.GalaxyCluster(\"Name of cluster\", cluster_ra, cluster_dec,\n",
+    "                                   cluster_z, galaxies)\n",
+    "\n",
+    "# Convert elipticities into shears for the members.\n",
+    "cluster.compute_tangential_and_cross_components(geometry=\"flat\")\n",
+    "print(cluster.galcat.colnames)\n",
```
combet, line 434: Please, remove `geometry='flat'`. The default is now to use the correct `curve` geometry, using astropy functionality.

combet (changes requested):
Hi Sheming and thank you. Going through the notebook, I realised `geometry=flat` should be avoided as we now have the accurate spherical description implemented using astropy. Simply remove that keyword and things should be computed using the spherical description.

examples/real_data.ipynb
shenmingfu: Thanks. I have removed it.

examples/real_data.ipynb
shenmingfu: Done.

shenmingfu:
The HSC PDR catalogs are proposed to be mirrored on NERSC soon.

combet:
Thanks Shenming - so as discussed during tag up last week, we'll move this PR along as it is and the notebook will be updated once the HSC data becomes available at NERSC.

@marina-ricci - given that you did the initial thorough review of this PR, could you go ahead and approve if everything else (but data access) as been resolved?

marina-ricci:
Looks good to me!

marina-ricci (approved):

combet (approved):
Thank you @shenmingfu and @marina-ricci

PR #460, opened 2021-11-08
Issue/443/stacked data

payerne:
This branch is dedicated to the implementation of WL weights for individual background galaxies, in the context of WL shear analysis (the implementation of the stacking strategy for shear profiles is in branch `#issue/127/write cluster ensemble`, the above branch name is not appropriated). It allows the user to specify the level of uncertainty to include in the individual galaxy weights (shape noise, shape measurement error, photometric redshifts) for both excess surface density and tangential shear. These functionalities are available in a new function `compute_galaxy_weights` that is added to the `GalaxyCluster` object. A new notebook gives an example in the context of photoz & shape measurement errors for background galaxies.


clmm/dataops/__init__.py
```diff
@@ -164,6 +165,143 @@ def compute_tangential_and_cross_components(
...
+        Indicates whether it is the excess surface density or the tangential shear
+    validate_input: bool
+        Validade each input argument
+
+    Returns
+    -------
+    w_ls: array
+        Individual lens source pair weights
+    p_background : array
+        Probability for being a background galaxy
```
combet, line 247: It is confusing to have that function also returns the probability of being a background galaxy. This should be done in a separate function, e.g. `compute_background_probability`.

examples/stacking_analysis/Test_compute_weights_GalaxyCluster.ipynb
```diff
@@ -0,0 +1,288 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "w_ls_photoz_shape = cl.compute_galaxy_weights(\n",
+    "    z_source='ztrue', pzpdf='pzpdf', pzbins='pzbins', \n",
+    "    shape_component1 = 'e1', shape_component2 = 'e2', \n",
+    "    shape_component1_err = 'e1_err', shape_component2_err = 'e2_err', \n",
+    "    add_photoz = True, add_shapenoise = True, add_shape_error = True, \n",
```
combet, line 165: `add` --> `use`

examples/stacking_analysis/Test_compute_weights_GalaxyCluster.ipynb
```diff
@@ -0,0 +1,288 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "w_ls_photoz = cl_2.compute_galaxy_weights(z_source='ztrue', pzpdf='pzpdf', pzbins='pzbins', \n",
+    "                        shape_component1 = 'e1', shape_component2 = 'e2', \n",
+    "                        shape_component1_err = 'e1_err', shape_component2_err = 'e2_err', \n",
+    "                        add_photoz = True, add_shapenoise = False, add_shape_error = False, \n",
```
combet, line 238: `add` --> `use`

clmm/galaxycluster.py
```diff
@@ -181,28 +209,133 @@ def compute_tangential_and_cross_components(
...
+        # computes weights
+        w_ls = compute_galaxy_weights(
+            self.z, cosmo, use_shape_noise=use_shape_noise,
+            is_deltasigma=is_deltasigma, validate_input=self.validate_input,
+            **cols)
+        if add:
+            self.galcat[weight_name] = w_ls
+        return w_ls
+
     def make_radial_profile(self,
```
combet, line 339: So we now have the options to compute weights, but I do not see this propagated as an option to perfom a weighted average in `make_radial_profile`. @payerne, I think you did that in the issue/127 (cluster ensemble) branch (see `make_radial_profile` in `dataops/__init__.py in branch 127`). Just to clarify - do we agree to postpone this when dealing with `issue/127/...`. @payerne, @m-aguena ?

combet (changes requested):
Thank you @payerne and @m-aguena. I have a few comments and questions below.

- So we now have the options to compute weights, but I do not see this propagated as an option to perfom a weighted average in make_radial_profile. @payerne, I think you did that in the `issue/127` (cluster ensemble) branch (see `make_radial_profile` in `dataops/__init__.py` in branch 127). Just to clarify - do we agree to postpone this when dealing with issue/127/.... @payerne, @m-aguena ?
- If you agree, I'd also add another option, where the source redshift to compute the weight is randomly drawn from the pdf (`use_photoz_rand`?). This is what was done in the DES mass calibration paper (McClintock et al., see snapshot below) and it would be useful to be able to repeat this approach.
Screenshot 2022-03-08 at 08 49 18

Some small notebook related points:

- [ ]  I think the `examples/stacking_analysis/Test_compute_weights...` notebook could be moved to the top example directory as it exemplifies the various options to compute those weights.
- [ ] As it is, it does not run. Please check my detailed comments below, that mostly have to do with using wrong option names.
- [ ] Please also add a markdown description of each options before the corresponding cells.
- [ ] I think printing out once the table to show the new columns added would be helpful to the user.

Also I'd like your  input on the following 2 noteboooks in `examples/stacking_analysis`:
- I don't think the covariance notebook should be pushed with that branch. Here the goal was to deal with galaxy weights only. And actually, the notebook, as is, does not run because of obsolete option names (probably from a old version of CLMM). Probably best suited for the cluster ensemble branch (`issue/127/...`)
- Same comment for the Example to fit the halo mass with the stacking approach. I think this should go and be updated in the "cluster ensemble" branch (`issue/127`), using the new weights functionality of this PR.

m-aguena:
Thanks @combet for a detailed review. I fixed the problems with outdated arguments in the notebooks and also moved `Covariance_matrix_for_stacked_analysis.ipynb` and `Example_fit_halo_mass_stacking_analysis.ipynb` to branch `issue/127/write_cluster_ensemblea`.

 With regards to propagating weights to `make_radial_profile`, it can be done in branch `issue/127/write_cluster_ensemble`.

@payerne can you address the other comments?

payerne:
Hi @m-aguena and @combet, I added comments on the example notebook relative to points mentioned by @combet above.

combet:
@m-aguena, so I approve this PR and locally checked that the tests and coverage are good. But the latest changes to GHA means we do not see that tests/coverage are OK on the interface, before merging to `main`. Is it the expected behaviour?

Issue #421, opened 2021-07-16
Changing alpha param in Einasto profile


marina-ricci:
At the moment, the Einasto profile is implemented in CCL (an NumCosmo?) in a way that prevents the user of changing the `alpha` parameter (in CCL this is computed internally (using the parameterization of Diemer & Kravtsov (arXiv:1401.1216)). 

I would like to include the possibility of changing the `alpha` parameter.

pennalima:
Hi @marina-ricci , in NumCosmo alpha is a parameter that can be fitted. For the moment, we do not consider any parametrization.

marina-ricci:
To develop : 

1. A warning printing the "alpha" value in case of Einasto profile for CCL and NC (maybe not the same dfault value)
2. Develop the feasibility to fit extra parameter in CLMM

combet:
Note - this is also linked to #402.

marina-ricci:
@pennalima : You can check Fig. 13 in this paper [https://academic.oup.com/mnras/article/441/4/3359/1209689](Dutton & Maccio 2014) for reference on the physical rang for alpha. Also note that alpha and the concentration are negatively correlated.

Issue #402, opened 2021-06-11
Implement Einasto and Herquist profiles from CCL


combet:
As shown in #398, it is possible to improve CCL's default integration to get enhanced precision, although, the results are quite sensitive to the choice of radial range and sampling. Once some configuration is agreed upon in #398, it should be implemented in CLMM.

combet:
I've implemented the profiles in `ccl.py`, following the integration recipe of the `validation_test` NB of branch 398. I checked that the `demo_theory` notebooks (functional or OO) are running, asking `Einasto` or `Hernquist` from the CCL backend.

TODO:
- [ ] get benchmark file from NumCosmo for testing(@pennalima ?)
- [ ] Implement corresponding tests.

combet:
This is also linked to #421

pennalima:
Hi @combet, benchmarks are done.

combet:
Hi @pennalima - thanks, but I don't see the new data files in the branch. Did you push those new files?

pennalima:
Hi @combet, 
In branch "example_Einasto", you can see the files 
CLMM/tests/data/numcosmo/radial_profiles_einasto.txt (radial_profiles_hernquist.txt and config_einasto_henrquist_benchmarks.json).

PR #461, opened 2021-11-18
Einasto + Hernquist for CCL backend + free Einasto slope parameter for NC backend

combet:
This originally only adressed #402. But because the Einasto slope `alpha` cannot be changed in CCL, I also implemented the possibility to change it for the NC backend in CLMM (so that NC could also pass the tests using the benchmarks generated for the CCL alpha value). So this PR also (partly?) addresses #421.

- Implements Einasto and Hernquist for the CCL backend
- Implements a user-defined Einasto alpha slope for the NC backend
- New `verbose` option in `theory` to print out the value of the Einasto `alpha` slope
- Updates notebooks to show the new functionality
- Adds benchmarks files for both profiles (+  notebook used to generate them, adapted from @pennalima's notebook).
- Updates CI tests to run on the Einasto and Hernquist profiles for CCL and NC backends.

~~Note that I've re-generated the benchmarks files only in a radial range where CCL's Einasto behaves OK, i.e, 0.01-10 Mpc (something strange is happening at larger radii that needs to be investigated). For Einasto and Hernquist, the tests currently ask for a loose relative tolerance only (1%), that will need to be made more stringent once the CCL results get more precise.~~

The strikethrough comment above has been solved by @hsinfan1996's updates on CCL numerical integration parameters.


hsinfan1996:
For pyccl.halos.profiles.HaloProfileEinasto and pyccl.halos.profiles.HaloProfileHernquist objects, I update the fftlog settings with 
```
pyccl.halos.profiles.HaloProfile.update_precision_fftlog(padding_lo_fftlog=0.0000001,
                                 padding_lo_extra=0.0000001,
                                 padding_hi_fftlog=1000000.,
                                 padding_hi_extra=1000000.,
                                 large_padding_2D=False,
                                 n_per_decade=100000,
                                 extrapol='linx_liny',
                                 plaw_fourier=-2,
                                 plaw_projected=-1)
```
Generally, Hernquist profile has better precision than Einasto profile (Sigma and DeltaSigma). In examples/Paper_v1.0/validation_tests.ipynb, the relative differences can be less than 2e-5 (2e-6 if set n_per_decade=1000000). However, in pytest, Einasto Sigma profile has max relative difference ~9e-3, and I cannot improve it.

hsinfan1996:
Right now CCL uses numerical integration when calculating Einasto density profile (3D), and the precision is not good. However, Einasto density profile can be done analytically, and Sigma and DeltaSigma can be calculated numerically based on the 3D profile (I think).

hsinfan1996:
In pyccl/halos/profiles.py, by replacing line 877 `norm, status = lib.einasto_norm(R_s, R_M, alpha, M_use.size, status)` with `norm = M_use/(np.pi*R_s**3*2**(2-3/alpha)*alpha**(-1+3/alpha)*np.exp(2/alpha)*gamma(3/alpha)*gammainc(3/alpha, 2/alpha*(R_M*a/R_s)**alpha))` (and reinstall CCL if needed), the max relative difference on Einasto 3D density profile will drop from 0.0078 to less than 1e-10 without touching any fftlog setting. The test for Einasto profile will pass with max relative difference of ~4e-09 on DeltaSigma profile. This calculation is what Colossus uses.

hsinfan1996:
It seems that my CCL (2.3.1.dev42+ga37cad61, the latest from Github) calculated a different value of einasto_alpha from the one in the notebook generate_einasto_hernquist_benchmarks_fromNC.ipynb. After updating the fftlog settings, the tests can pass with reltol=5e-5. It can be improved, but the fftlog calculation will take a lot more time.

combet:
Thank you for the update @hsinfan1996. The precision issue I had was then most likely due to me not using the right alpha value for the benchmarks (I might have rounded it).

hsinfan1996:
Setting `padding_lo_fftlog` to a lower value does not help with the overall precision, but just to make sure Hernasto Sigma profile does not deviate from NumCosmo values too much at large radii (>10 Mpc).

m-aguena:

combet:
@m-aguena, @marina-ricci, @pennalima - I've finished implementing the missing tests for that PR. It's now ready for review.

m-aguena:
@combet can you merge master into it and fix the conflicts? thanks

combet:
@m-aguena. Merge conflicts are now fixed. After fixing the merge conflicts, the test for Einasto were failing and it took me some time to find what the issue was. 

Actually, the way CCL computed the value of the Einasto slope depends on the matter power spectrum. The PS default was changed from `bbks` to `eisenstein_hu` for the 2-h term implementation. So the benchmark values for the Einasto profiles actually had to be recomputed. All is good now.

clmm/theory/func_layer.py
```diff
@@ -661,8 +713,10 @@ def compute_magnification(r_proj, mdelta, cdelta, z_cluster, z_source, cosmo, de
             halo_profile_model=halo_profile_model, massdef=massdef, delta_mdef=delta_mdef)
         gcm.set_concentration(cdelta)
         gcm.set_mass(mdelta)
+        if alpha_ein is not None:
+            gcm.set_einasto_alpha(alpha_ein)
```
m-aguena, line 717: Can you add a test for this line?

m-aguena (changes requested):
Overall it seems good, just a couple of comments I left on the code.

clmm/theory/parent_class.py
```diff
@@ -140,7 +175,31 @@ def _set_concentration(self, cdelta):
         r""" Actuall sets the value of the concentration (without value check)"""
         raise NotImplementedError
 
-    def eval_3d_density(self, r3d, z_cl):
+    def _validate_input(self, in_val, vmin, err_msg='value %s <= vmin'):
```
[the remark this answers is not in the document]
combet: I removed them - indeed, not used anywhere

clmm/theory/func_layer.py
combet: Done.

clmm/theory/ccl.py
```diff
@@ -55,23 +56,24 @@ def __init__(self, massdef='mean', delta_mdef=200, halo_profile_model='nfw',
...
+        self.hdpm_dict = {'nfw': ccl.halos.HaloProfileNFW,
+                          'einasto': ccl.halos.HaloProfileEinasto,
+                          'hernquist': ccl.halos.HaloProfileHernquist}
         self.cosmo_class = CCLCosmology
-        # Uncomment lines below when CCL einasto and hernquist profiles are stable
-        # (also add version number)
-        # if version.parse(ccl.__version__) >= version.parse('???'):
-        #    self.hdpm_dict.update({
+        # Uncomment lines below when CCL einasto and hernquist profiles are stable (also add version number)
+        #if version.parse(ccl.__version__) >= version.parse('???'):
```
m-aguena, line 64: Are we always going to require am installed CCL version with Einasto and Hernquist profiles? It this why is this line still commented?

m-aguena (commented):
Overall all seems good, I just left one comment

clmm/theory/ccl.py
hsinfan1996: @combet @m-aguena For your information, the halos modules were added to CCL in version 2.0.2.dev84. An official release with Einasto and Hernquist profiles would be 2.1.0.

combet: Currently pip-installable ccl release is 2.3.0, so I think it's OK to request >=2.1.0

combet:
@m-aguena - done as suggested above. Now clearly requesting ccl >=v2.1.0 and commented out lines are removed.

m-aguena (approved):

Issue #465, opened 2021-12-03
Add functions to compute <beta_s> and <beta_s^2>


m-aguena:
The implementation for reduced tangential shear of Weighting the Giants in PR #425 expects `beta_s_mean` and `beta_s_square_mean` as an input. Add functions to compute those.

marina-ricci:
Preliminary resolution in draft PR #467

PR #425, opened 2021-08-08
Update theory prediction of g_t to account for a galaxy source distribution

shenmingfu:
#397 
I added `z_src_model='weighing_the_giants_b'` in `compute_reduced_tangential_shear()` in `clmm/theory/func_layer.py`.
Later I will try to add it in `ccl.py`, `cluster_toolkit.py`, and `numcosmo.py` as well.

PR #467, opened 2021-12-03
Issue #465 add beta functions

marina-ricci:
PR to address issue#465.


combet:
Hi @marina-ricci - Looking at the first commit of this PR, I see references to Einasto and Hernquist profiles, which shouldn't be there I think. Did you started branch 465 from `main`?

Also, I just merged PR #425 into `main`, so whatever the situation is, make sure to merge `main` again.

marina-ricci:
> Hi @marina-ricci - Looking at the first commit of this PR, I see references to Einasto and Hernquist profiles, which shouldn't be there I think. Did you started branch 465 from `main`?
> 
> Also, I just merged PR #425 into `main`, so whatever the situation is, make sure to merge `main` again.

Yes, there were some strange commits in main, I corrected that now but I don't think it matters.

eduardojsbarroso:
Hi @marina-ricci , I have a question about the betas_mean function. I get that the variable times its density distribution divided by the normalization factor is the standard way of computing the mean value of a random variable. But why didn't you use the definition in equation (13) from [https://ui.adsabs.harvard.edu/abs/2018MNRAS.474.2635S/abstract](Shraback et al)? And if I do continue with the equation that you proposed, should we expect that the user provide the redshift density probability or should I implement a function to compute it, as in equation (3.8) from [https://articles.adsabs.harvard.edu/pdf/1997A%26A...318..687S](Carolin Seitz and Peter Schneider)?

marina-ricci:
(Answered live). Eq. 13 is for the case when you have individual redshift for each source galaxy. We should allow the user to give any functional form for the redshift distribution.

aimalz:
> (Answered live). Eq. 13 is for the case when you have individual redshift for each source galaxy. We should allow the user to give any functional form for the redshift distribution.

A good way to do that is to allow the user to pass in a [qp](https://github.com/LSSTDESC/qp) object (ensemble with a single row).

marina-ricci:
@eduardojsbarroso thanks for the PR! I think it would be better to work with `scipy.integrate.quad` than with `np.nansum`.

eduardojsbarroso:
@marina-ricci Thanks for the suggestion. It is implemented now. I had to change the arguments in the beta funcions (since the integration integrates over the first argument), and had to create a temporary integrand in the mean functions. Since we are computing a product of funcions, where each function takes a different amount of arguments, I can't use:
`scipy.integrate.quad(lambda x: B(x,y,z) * z(x)), xmin, xmax, args = (y,z)`

Let me know if you want me to think in another solution or if this is good enough.

eduardojsbarroso:
@marina-ricci  Also, I don't know why the build failled to install coverall

clmm/utils.py
```diff
@@ -459,3 +459,192 @@ def validate_argument(loc, argname, valid_type, none_ok=False, argmin=None, argm
...
+    
+    Returns
+    -------
+    float
+        Mean square value of the geometric lensing efficicency ratio.
+    """ 
+    if pdz == None:
+        alpha, beta, redshift0 = 1.24, 1.01, 0.51
+        def pdz(z):
+            return (z**alpha)*np.exp(-(z/redshift0)**beta)
```
hsinfan1996, line 642: Do we want to move (or import?) _chang_z_distrib from support/mock_data.py to utils.py so we don't have to define it again here?

combet: I agree we want them in one place only. But actually, I'd go the other way around. Have the pdf defined in `utils` and import them to `mock_data`.

hsinfan1996: Agree. Do we still keep it as a private function? or do we want to make it a public function?

combet: I don't have a strong opinion, but can't think of any objection to making it public.

clmm/support/mock_data.py
```diff
@@ -332,12 +284,8 @@ def _generate_galaxy_catalog(cluster_m, cluster_z, cluster_c, cosmo, ngals,
...
 
     posangle += 0.5*np.pi
-    if np.iterable(posangle):
-        posangle[posangle > np.pi] -= 2*np.pi
-        posangle[angsep == 0] = 0
-    else:
-        posangle -= 2*np.pi if posangle > np.pi else 0
-        posangle = 0 if angsep == 0 else posangle
+    posangle[posangle > np.pi] -= 2*np.pi
+    posangle[angsep == 0] = 0
```
hsinfan1996, line 288: NOTE: This is the leftover code when we change from flat-sky to Astropy SkyCoord.
galaxy_catalog['ra'] and galaxy_catalog['dec'] are `astropy.table.column.Column` objects, so c_gal cannot return a scalar. `c_cl.isscalar` can be True when instantiated with both cluster_ra and cluster_dec being scalars. But after the operation `c_cl.position_angle` with non-scalar c_gal, `c_cl.isscalar` is False and therefore `np.iterable(posangle)` is always True.

marina-ricci (commented):
Change function names for consistency: 
- compute_beta_mean
- compute_beta_s_mean 
Clarify the "compute_B_" functions documentation : 
- zmax = increase default value to e.g. 10
- pdz : "argument function" (remove "1")
 - delta_z_cut => clarify explanation (mention that it is not used if zmin is defined).

clmm/theory/parent_class.py
```diff
@@ -371,21 +381,21 @@ def eval_reduced_tangential_shear(self, r_proj, z_cl, z_src, z_src_model='single
...
-                raise ValueError("beta_s_mean or beta_s_square_mean is not given.")
-            else:
-                z_source = 1000. #np.inf # INF or a very large number
-                gammat = self._eval_tangential_shear(r_proj, z_cl, z_source)
-                kappa = self._eval_convergence(r_proj, z_cl, z_source)
-                gt = (1. + (beta_s_square_mean / (beta_s_mean * beta_s_mean) - 1.) * beta_s_mean * kappa) * (beta_s_mean * gammat / (1. - beta_s_mean * kappa))
+                beta_s_mean = compute_Bs_mean (z_cl, 1000., self.cosmo, pdz=pdz)
+                beta_s_square_mean = compute_Bs_square_mean (z_cl, 1000., self.cosmo, pdz=pdz)
+            z_source = 1000. #np.inf # INF or a very large number
+            gammat = self._eval_tangential_shear(r_proj, z_cl, z_source)
```
marina-ricci, line 396: Mention z_source beforehand so we can have z_inf =  z_source in the functions instead of hardcoded 1000.

hsinfan1996:
I am actually not clear what "argument function" is. But it seems to me that `pdz` can  be described as a "one-argument function", "one-parameter function" or maybe more formally "univariate function".

clmm/theory/parent_class.py
```diff
@@ -334,15 +334,25 @@ def eval_reduced_tangential_shear(self, r_proj, z_cl, z_src, z_src_model='single
                     (Applegate et al. 2014; https://arxiv.org/abs/1208.0605) to evaluate tangential reduced shear;
                 * `schrabback18`: use the equation (12) in Cluster Mass Calibration at High Redshift \
                     (Schrabback et al. 2017; https://arxiv.org/abs/1611.03866) to evaluate tangential reduced shear;
-                
+        
+        pdz: one parameter function
+            The probability density distribution of the redshift. This function takes /
+            the redshift as its only parameter and returns its probability. It is /
```
hsinfan1996, line 340: To me, a PDF should not return a probability if we are not taking the integral of it. But I am also not sure how to best describe this part.

hsinfan1996:
> > I am actually not clear what "argument function" is. But it seems to me that `pdz` can  be described as a "one-argument function", "one-parameter function" or maybe more formally "univariate function".
> 
> @hsinfan1996 That's what I wanted to say with 1 argument function before, but maybe the format wasn't clear. I will rename it.
> 
> And for the pdz, the integral of a PDF is the cumulative function, or am I wrong? For me this is indeed a probability.

Yes, you are right. By integrating PDF in a certain way we get CDF. But evaluating a CDF at a point gives us a probability while evaluating a PDF at a point does not (probability is 0).

combet:
Thanks @eduardojsbarroso, @marina-ricci and @hsinfan1996. I started reviewing this PR but noticed that the new functionalities were not exemplified in a notebook. I think it would be good to have at least one example of the non 'single_plane' case in `demo_theory_functionality.ipynb`.

clmm/theory/parent_class.py
```diff
@@ -426,7 +426,7 @@ def _eval_convergence(self, r_proj, z_cl, z_src):
         return sigma/sigma_c
 
     def eval_reduced_tangential_shear(self, r_proj, z_cl, z_src, z_src_model='single_plane',
-                                      beta_s_mean=None, beta_s_square_mean=None):
+                                      beta_s_mean=None, beta_s_square_mean=None, z_distrib_func=None):
```
combet, line 429: With the new non-single-plane options, should `z_src` be made an optional argument, with `None` a valid option? In that case, we won't have individual `z_src` values to provide.

Linked to that, and if agreed with, in the argument validation step (line 471-474):
- it should be extended to check the types of the new `beta` parameters;
- If `z_src` becomes an optional parameter, with `None` as a valid value, then the validation line should become `validate_argument(locals(), 'z_src', 'float_array', argmin=0, none_ok=True)`.

eduardojsbarroso: I agree with making `z_src` an option variable. Maybe for the test it could be something like
````
if z_src_model=='single_plane':
else:
    validate_argument(locals(), 'z_src', 'float_array', argmin=0, none_ok=True).
````
This way `None` won't be accepted if the function is called with the `single_plane` option.

**Comment by marina-ricci** [comment]
@eduardojsbarroso I'm working on implementing the new functionalities  in the demo notebook (just so we don't do the same thing :))

File: clmm/theory/parent_class.py [hunk 6]
```diff
@@ -425,7 +425,7 @@ def _eval_convergence(self, r_proj, z_cl, z_src):
         sigma_c = self.eval_critical_surface_density(z_cl, z_src)
         return sigma/sigma_c
 
-    def eval_reduced_tangential_shear(self, r_proj, z_cl, z_src, z_src_model='single_plane',
+    def eval_reduced_tangential_shear(self, r_proj, z_cl, z_src=None, z_src_model='single_plane',
```
eduardojsbarroso, line 428: @combet If I make the same change at `compute_reduced_tangential_shear` in `func_layer.py`, I will have to change the order of the parameters, since the function is
````
compute_reduced_tangential_shear(
        r_proj, mdelta, cdelta, z_cluster, z_src, cosmo,
        delta_mdef=200, halo_profile_model='nfw', massdef='mean',
        z_src_model='single_plane', beta_s_mean=None, beta_s_square_mean=None,
        z_distrib_func=None, validate_input=True):
````

I will have to change such as the new order  will be `cosmo, z_src=None`, and thus we must change the examples in the notebook where the functions are called with a parameter order. Should I change the function and check all the notebooks?

combet: Hi @eduardojsbarroso - I think this is OK to change but would like @m-aguena and @marina-ricci's input as this means the API will not be backward compatible. Users currently using `compute_reduced_tangential_shear` will have to change the argument ordering for it to work. I don't think this is a major issue at this stage if properly tagged, but would be best to have agreement on this.

marina-ricci: Comment : We will implement that change in a separate PR.

combet (approved):
All looks good (we leave the question of z_src as optional or not parameter to another issue).

PR #473, opened 2022-01-18
Issue/470/mass conversion

hsinfan1996:


hsinfan1996:
@combet In cosmology/numcosmo.py, most of the NumCosmo functions only accept argument types of float and size-1 array. Right now validate_argument for the corresponding methods in cosmology/parent_class.py allows float_array, but there will be errors for NC backend if there is more than one element in the array. Do we want to vectorize all the NC functions or change the validation type?

combet:
Hi @hsinfan1996 - we should make sure that both CCL and NumCosmo backend are able to handle the same type of arguments, and being able to pass an array is definitely a plus. I think this was discussed on the NumCosmo side at some point, @vitenti @pennalima? Looks like it's time to revive that discussion

hsinfan1996:
> Hi @hsinfan1996 - we should make sure that both CCL and NumCosmo backend are able to handle the same type of arguments, and being able to pass an array is definitely a plus. I think this was discussed on the NumCosmo side at some point, @vitenti @pennalima? Looks like it's time to revive that discussion

I see. Both Astropy and CCL can handle an array as an argument correctly.

m-aguena:
@hsinfan1996 can you check if the functions added in cosmology here are not redundant with those in PR #459 ? Thanks

hsinfan1996:
> @hsinfan1996 can you check if the functions added in cosmology here are not redundant with those in PR #459 ? Thanks

@m-aguena Thanks for reminding me. From what I see, the functions are different.

hsinfan1996:
@m-aguena Thank you. It is working as expected. Does that mean we can remove profiles.py? When a Modeling() object is initialized, delta_mdef should be set. However, CCL and NC backend require an additional step `model.delta_mdef=200`. Should we do `self.delta_mdef=delta_mdef` during  `self.set_halo_density_profile`?

m-aguena:
@hsinfan1996 yes, profiles.py can be removed now. With regards to `delta_mdef`, yes I also think we should set it up inside `set_halo_density_profile`, but I also think we can restructure this function a bit to have modeling more organized. I am just waiting for some confirmation in issue #513

clmm/theory/generic.py
```diff
@@ -67,3 +69,63 @@ def compute_magnification_bias_from_magnification(magnification, alpha):
...
+                                   massdef='mean', delta_mdef=200,
+                                   halo_profile_model='nfw', alpha=None):
+    rdelta = compute_rdelta(mdelta, redshift, cosmo, massdef, delta_mdef)
+    # Check halo_profile_model
+    if halo_profile_model=='nfw':
+        prof_integ = lambda c: np.log(1. + c) - c/(1. + c)
+    elif halo_profile_model=='einasto':
+        if alpha is None:
+            raise ValueError('alpha must be provided when Einasto profile is selected!')
+        prof_integ = lambda c: gamma(3./alpha)*gammainc(3./alpha, 2./alpha*c**alpha)
```
hsinfan1996, line 91: A side note: CCL uses numerical integration to calculate this factor (for the density profiles), while both CT and NC use this analytic form. I don't think it will affect anything, but I would like to bring it up.

hsinfan1996: Yes. I am also wondering why CCL chose to use numerical integration when the analytic expression was available.

m-aguena:
@hsinfan1996 I implemented the tests you had made before for the mass convesion, but it is failing with einasto. Can you take a look at the value there? Thanks

hsinfan1996:
> @hsinfan1996 I implemented the tests you had made before for the mass convesion, but it is failing with einasto. Can you take a look at the value there? Thanks

Fixed. The check failed because of a very minor error from compiling the documentation.

hsinfan1996:
> 
Thanks for trying out the notebook. 
`examples/Compare_NFW_critical_massdef.ipynb` was from PR #495.
How do you calculate the enclosed masses from def2? Could you try `alpha=alpha1` in the last cell when calculating enclosed masses of def2 with `clmm.compute_profile_mass_in_radius`? The Einasto alpha is not preserved after conversion and cannot be set in the CCL backend.

combet:
@hsinfan1996 - 
- OK for `examples/Compare_NFW_critical_massdef.ipynb` - I missed it at the time I think.
- I think I propagated the CCL alpha properly throughout the notebook and still the same. I'm not sure I understand what you mean by "alpha is not preserved after conversion". Do you mean you convert M,c and alpha for Einasto? In addition, here's what I get for Hernquist, only changing the halo profile model in cell 4 (and 5). Once again, the relative difference is quite large. Is this expected ? Could you try to reproduce these as your end, starting from your notebook. I've I'm misusing something somewhere I'd like to understand where.
Screenshot 2022-06-10 at 09 48 01

hsinfan1996:
@combet I am able to reproduce the results. In `clmm.compute_profile_mass_in_radius`, `halo_profile_model` has to be specified ('einasto'` or `'hernquist') because the default value is `'nfw'`. Sorry I missed this part in the notebook. I will add it to the last cell.
About einasto alpha not preserved:
if I create another Modeling object for def2 with
```
nfw_def2 = clmm.Modeling(massdef=massdef2, delta_mdef=delta_mdef2, halo_profile_model='einasto')
nfw_def2.set_mass(M2)
nfw_def12set_concentration(c2)
nfw_def2.set_cosmo(cosmo)
```
I got `nfw_def2.get_einasto_alpha(z_cl)/nfw_def1.get_einasto_alpha(z_cl)-1` about 0.0005.

combet:
Ah OK, I understand ! Great, it seems all works well. I'll continue with the review then !

combet:
@hsinfan1996, @m-aguena - as just mentioned at tag up, I just pushed the new notebook updated with the Einasto and Hernquist examples as well. Could you please add the analytical expressions of the enclosed mass for the various profiles, in particular so we see the need to provide `alpha` for the Einasto profile
- [ ] in the notebook
- [ ] in the docstrings of the `compute_profile_mass_in_radius` / `eval_mass_in_radius`

hsinfan1996:
> @hsinfan1996, @m-aguena - as just mentioned at tag up, I pushed the new notebook updated with the Einasto and Hernquist examples as well. Could you please add the analytical expressions of the enclosed mass for the various profiles, in particular so we see the need to provide `alpha` for the Einasto profile
> 
> * [x]  in the notebook
> * [x]  in the docstrings of the `compute_profile_mass_in_radius` / `eval_mass_in_radius`
>   Also
> * [x]  Rename the notebook to `demo_mass_conversion.ipynb`, without mentioning NFW in the title as it now does the 3 profiles.

@combet Done. In the notebook, I don't know how to improve the typesetting of the equations, but they are looking okay right now.

combet (approved):
Thanks @hsinfan1996 and @m-aguena. All looks good!

Issue #477, opened 2022-01-24
Update mock data to allow for a cluster not centered in (0,0)


combet:
At the moment, `mock_data` only supports a cluster centered in (0,0) as the background galaxies are drawn in a field centered in (0,0). Need to update so that
- [x] Pass `ra_cl` and `dec_cl` as keywords of `generate_galaxy_catalog`. If `None`, keep (0,0) as the default.
- [x] Propagates this to `_draw_galaxy_positions`
- [x] Update `_draw_galaxy_positions` to uniformly draw positions on the sphere and not in the flat sky as is currently the case.
- [x] Update the corresponding demo notebook to highlight this functionality

Issue #468, opened 2021-12-03
Problem with "validate_input" in example notebooks


marina-ricci:
Since we added the "validate_input" functionality to functions, various example notebooks do not run properly anymore. 
Specifying "validate_input"=False on the notebooks will work. 
@m-aguena this raise the question of the current use of "validate_input". If we have to turn it off for most of the science application, maybe it is worth updating it.

hsinfan1996:
Scipy.optimize functions used by fitters and samplers will convert arguments to np.ndarray with np.copy(). It would be great if the ndarray can be converted back with numpy.ndarray.item() or other method even when validate_input=True.

marina-ricci:
That is what fails the test in the notebooks, but there are also cases where you may want to input, e.g. an array of mass. Sure in this case you could put "validate_input"= False, but it is not an invalid input per se. Maybe the validate_input should be present only in the GalaxyCluster object but not on the individual functions.

hsinfan1996:
In the case of an array of mass, I think the validation will pass if valid_type is set to float_array (also int_array if they are integers).

m-aguena:
@marina-ricci in the current version of the code we have these functions set up to get `float` mass. If we want to have the possibility of using array of masses you just have to use `'float_array'` instead of just `float` in the `validate_argument` function and also update the documentation accordingly.

PR #478, opened 2022-01-31
Issue/477/mock cluster position

combet:
Allows for mock data to generate a galaxy field around a cluster not centered in (0,0). Adresses #477 
- Now, `cluster_ra` and `cluster_dec` are keywords of `mock.generate_galaxy_catalog`. If not specified, the default remains (0,0)
- All notebooks have been modified accordingly

In terms of procedure:
- galaxies are first drawn around a cluster in (0,0), in the flat sky approx 
- using astropy, separation and position angles are then computed with respect to (0,0)
- use astropy to generate new coordinates, using these separation and position angles, but around a new cluster position


hsinfan1996 (approved):

hsinfan1996:
#468 is present in Example1, though we might not fix it here. So this is just a reminder only.

combet:
@hsinfan1996 - thanks for the reminder. Yes, we need to deal with this sooner rather than later, but it should indeed be done elsewhere I think.

Issue #464, opened 2021-12-03
Updating Example 3


shenmingfu:
Details at 
https://github.com/LSSTDESC/CLMM/pull/425#issuecomment-953832367

shenmingfu:
In section "Radially binning the data" we need to mention that source redshifts are also binned -- so in the radial profile (table) each z value is the mean in that radial bin (details in `dataops/make_radial_profile` and `utils/compute_radial_averages`). 

Then in section "Create the reduced tangential shear models" we can define a new function that uses `beta_s_mean` and `beta_s_square_mean`, similar to the ones we showed in the notebook for paper v1.0 (gt_and_use_case.ipynb) and the notebook for real data. 
Its arguments include logm, profile, catalog (for getting `.galcat['z']` to build `beta_s`) ; they are similar to the ones of `model_reduced_tangential_shear_zdistrib()` there.

Then we can use this new function in later sections; the usage will be similar to the `model_reduced_tangential_shear_zdistrib()` function.

PR #475, opened 2022-01-20
Issue/463/Added 'schrabback' option to reduced shear

eduardojsbarroso:

Issue #397, opened 2021-05-25
Update theory prediction of g_t to account for a galaxy source distribution ?


combet:
So far, the theory allows to predict the shear for a given source redshift (single source plane option). But we can use the overall redshift distribution of the sources to predict the corresponding shear, following:

See [WtGIII](https://arxiv.org/abs/1208.0605) (color-cut method).

![Screenshot 2021-05-25 at 10 10 37](https://user-images.githubusercontent.com/20227809/119462844-75189000-bd41-11eb-8a7e-e8cf698a3621.png)

where
![Screenshot 2021-05-25 at 11 35 00](https://user-images.githubusercontent.com/20227809/119475227-3ee10d80-bd4d-11eb-8422-43db63344db5.png)

An example of this was coded up in [this notebook](https://github.com/LSSTDESC/CLMM/blob/issue/335/nb_for_paper1.0/examples/Paper_v1.0/mass_bias_from_models.ipynb) for the v1.0 paper. But we should make this a proper option in the main code base.

This is linked to #130

m-aguena:
I  think a way to implement this, and also prepare to use full pdf would be to pass different quantities with the argument `z_source`  depending on the value of `z_src_model` in `clmm.theory.compute_reduced_tangential_shear`. Something like

1. `z_src_model='single_plane'` -> `z_source`  is one single value representing all sources (float)
2. `z_src_model='radial_values'` -> `z_source`  is values of readshifs in each radial bin (array with same size of `r_proj`)
3. `z_src_model='weighing_the_giants'` -> `z_source` is the redshift of individual sources (array with size = n sources)
4. `z_src_model='weighing_the_giants_b'` -> `z_source`  is <beta_s> (float)
5. `z_src_model='photoz_pdf'` -> `z_source`  is the full photo-z pdf (2d array)

The code already can run cases 1 and 2, but the documentation is not expicit about it.

shenmingfu:
I was about to ask whether we need to solve an effective `z_source` value from <beta_s> using (2), then I realized <beta_s^2> could give a different effective `z_source`. Maybe it's easier to use <beta_s> and <beta_s^2> as the inputs of the function?

shenmingfu:
When I looked at the code structure, I realized the issue is a little more complicated than I thought. Should we try modifying the `def compute_reduced_tangential_shear(...)` in `CLMM/clmm/theory/func_layer.py`?

[...]

PR #479, opened 2022-02-03
Example3 update

shenmingfu:
Issue #464


shenmingfu:
The applegate14 model has been added and tested in example3. I will add schrabback18 after it's merged into the main.

shenmingfu:
Now the "schrabback18" option (#475) is also added into the notebook.

marina-ricci:
- Header text : Correct typos (e.g. "the Chang. (2013)", "yield a bias mass", " el at."). Modify the sentence " add a model using the method from Applegate el at. (2014)" to explain what method it is (e.g. "method to account for the redshift distribution from Applegate et al.")

marina-ricci (changes requested):
- Header text :
- [x]  Correct typos (e.g. "the Chang. (2013)", "yield a **bias** mass", " **el at.**"). 
- [x] Modify the sentence " add a model using the method from Applegate el at. (2014)" to explain what method it is (e.g. "method to account for the redshift distribution from Applegate et al.")

- Typo in the notebook : 
- [ ] refer to "Chang et al. 2013" and not "the Chang. (2013)", 
- [ ]  "the corresponding model is simply given by **the fucntion** below:"

- [ ] Add comments about accounting for the redshift distribution using the "Applegate" method under the profiles figure.

marina-ricci:
Thank you @shenmingfu for the notebook!
Please when you make a PR add a bit of context and list what you changed, so it is easier to review, without having to dig in the issue's history. Here you changed 2 notebooks, whereas I think you only need to modify "Example3". Another point is that this notebook will have to be updated after PR #465 is merged.

shenmingfu:
Hi @marina-ricci , Thank you for the review. I have adjusted the texts. The previous edits include adding comments to make the text more clear, adding functions to use applegate14 and schrabback18 methods, and adding commands to run those new functions (making shear profiles, fitting masses, and making plots).
I'm not sure why Example4 were also changed. But it looks we can adopt those changes? Since they switch from environment `desc` to `clmm-env`. If not, I can remove those changes.

m-aguena:

shenmingfu:
@marina-ricci Sorry for my delay. I'm planning to use the newly defined functions `compute_beta_s_mean` and `compute_beta_s_square_mean` etc. from the new `clmm/utils.py`. In my current Example3, I defined similar functions and quantities in the notebook, so I will replace them with the ones in `clmm/utils.py`. Do you think that will be the solution? 

However, that will require updating files like `clmm/utils.py` in this branch. Currently, only Example3 is the file that has been changed (the modification of Example 4 can be skipped). Should I modify those files in this branch as well? If so, could you teach me how to do that properly (a lot of files were updated in issue #465 and PR #467)? Otherwise, can we merge this branch first, so those files will be updated, then I adjust the Example 3 in a new branch? 

Thanks.

shenmingfu:
@marina-ricci I have updated the notebook using the updated function `compute_reduced_tangential_shear()`, and now `beta_s_mean` and `beta_s_square_mean` are computed internally; we don't need to provide them in the notebook. The PR is ready for review. While updating the notebook, I noticed a typo in the original definition of beta, so I also updated that.

shenmingfu:
@marina-ricci Thanks for your feedback. The changes in Example4 are the python version and environment. We can skip that.
I have updated `z_distrib_func` in those functions in Example3.
The reduced shear formulae of Applegate14 (https://github.com/LSSTDESC/CLMM/issues/397#issue-900556489) and Schrabback18 (https://github.com/LSSTDESC/CLMM/pull/425#pullrequestreview-793004842) are identical if we skip the Taylor-expansion terms higher than the first order of kappa_inf (i.e. f(beta_s)*(kappa_inf)^n, n>=2). Here we are using the same redshift distribution (the default Chang et al. 2013). So they are close when kappa_inf is small (e.g. when the radial distance is not too small). For the noisy data, the error bars of fitted masses derived from the two methods are slightly different (~0.3%).

marina-ricci (approved):

marina-ricci:
@shenmingfu looks good to me! I'll merge the PR.

shenmingfu:
@marina-ricci Thank you.

PR #482, opened 2022-02-11
clarify mention to cross shear

marina-ricci:
This PR solves issue #476  : I made small modifications to the notebook showing the miscentering effects.


combet (changes requested):
Thanks @marina-ricci. Two minor things I noticed:
- [ ] This sentence needs updating as the clusters is not in (0,0) anymore. _"We want to load this mock data into several CLMM cluster objects centered between (-0.2, -0.2) - (0.2, 0.2) degrees and save them in a list. The user can change the number of cluster centers if desired. We set the first center to (0,0) for comparison reasons, which corresponds to a == 0"_
- [ ] Outputs need to be cleared

marina-ricci:
@combet thank you! Yes, please take care if you have hands on it already and merge.

combet (approved):
All looks good!

PR #486, opened 2022-02-18
Issue/468/validate input in nb

m-aguena:
Fixed the problem of issue #468 on the notebook itself.


combet (approved):
Thanks @m-aguena - this looks good.

Note that I took the opportunity to put the mock cluster in the flat_sky NB not in (0,0) for display purpose (following the implementation of that option last week, a cluster in (0,0) will have some ra~0 and some ra~360, which makes displaying the source field awkward). I had missed that notebook last week when I propagated that option through the notebooks.

PR #488, opened 2022-02-20
Issue/487/update GitHub actions

hsinfan1996:


hsinfan1996:
Still need to update README regarding the minimum Astropy version.

combet:
@hsinfan1996 - could you comment on why there's a need for `Build and Check Minimum`. Excluding coverage, there is now 4 checks to pass, instead of 2 before. That seems like a lot of redundancy?

hsinfan1996:
> Thanks @hsinfan1996 - could you comment on why there's a need for `Build and Check Minimum`. Excluding coverage, there is now 4 checks to pass, instead of 2 before. Is this absolutely necessary?

There are only two checks actually. Once a pull request is open, any commit will be checked twice since we request checks to be performed on both push and pull request. They are redundant I would say, in the way that we only need checks when a commit is pushed. `Build and Check Minimum` is not absolutely necessary, but since we are adding new functionalities, I thought it would be useful if we could check the minimum requirement stated in the readme to ensure backward compatibility.

hsinfan1996:
Now there are only two checks to run after a pull request is created. We can also make the minimal check significantly faster by installing and testing CCL backend only.

m-aguena:

hsinfan1996:
Checks can complete within ~4 min.

m-aguena (approved):
I all looks good

PR #495, opened 2022-03-11
Issue/493/CT critical mass definition

hsinfan1996:
Shown in examples/Compare_NFW_critical_massdef.ipynb, the precision of the NFW profiles given by CT is limited by cosmology-related calculations.


hsinfan1996:
I don't think it is necessary to keep the new notebook.

hsinfan1996:
Do I need to add _get_E2 to CCL and NC backends (and get_E2 to parent_class) as well since I added it to CT backend?

m-aguena (changes requested):
Overall it looks good. @hsinfan1996 can you update the docstrings at `func_layer` and remove the `not in cluster_toolkit;` comments for the `massdef` arg?

Another thing, in the notebook I am getting the error
```gi.repository.NumCosmo' object has no attribute 'halo_density_profile_nfw_class_set_ni```.
Is this related to a newer version of numcosmo? I had just updated mine

hsinfan1996:
> Overall it looks good. @hsinfan1996 can you update the docstrings at `func_layer` and remove the `not in cluster_toolkit;` comments for the `massdef` arg?
> 
> Another thing, in the notebook I am getting the error `gi.repository.NumCosmo' object has no attribute 'halo_density_profile_nfw_class_set_ni`. Is this related to a newer version of numcosmo? I had just updated mine

Thank you @m-aguena. I fixed the docstrings and added some tests. About `halo_density_profile_nfw_class_set_ni`, it has already been in Paper_v1.0/validation_tests.ipynb. I did not get any error from NumCosmo, and I am afraid I cannot solve this problem.

m-aguena (changes requested):
Overall it is good, but can you clean the notebook?

m-aguena:
@hsinfan1996 I updated my ccl and the notebook worked. But I still think it might be worth adding the version check so the user can know where the problem comes from.

hsinfan1996:
> 

@m-aguena Yes, but can you specify which cells to clean? I had removed some unneeded lines when I push the notebook.

hsinfan1996:
> @hsinfan1996 I updated my ccl and the notebook worked. But I still think it might be worth adding the version check so the user can know where the problem comes from.

I fix the h_over_h0 by calling it in a different way. I tested CCL 2.1.0, 2.2.0 and 2.3.1 and they all worked fine.

m-aguena:
> > 
> 
> @m-aguena Yes, but can you specify which cells to clean? I had removed some unneeded lines when I pushed the notebook.

All cells, only clean notebooks should be added

hsinfan1996:
> 

> > > 
> > 
> > 
> > @m-aguena Yes, but can you specify which cells to clean? I had removed some unneeded lines when I pushed the notebook.
> 
> All cells, only clean notebooks should be added

I see. Done.

m-aguena (approved):

PR #504, opened 2022-04-15
Issue/490/standard ra

m-aguena:
Make ra in cluster and galcat to be between [0, 360[


combet (approved):
Thanks @m-aguena, all looks good!

PR #510, opened 2022-05-06
Issue/507/ccl_eval_da

hsinfan1996:
Make _eval_da_z1z2 for CCL backend behave the same way as for the other two backends. Though many lines of code are added, it is still faster than the original numpy.vectorize implementation.

### DIFF
--- a/clmm/__init__.py
+++ b/clmm/__init__.py
@@ -12,4 +12,4 @@ from .theory import (
 )
 from . import support
 
-__version__ = '1.1.9'
+__version__ = '1.1.10'
--- a/clmm/cosmology/ccl.py
+++ b/clmm/cosmology/ccl.py
@@ -86,7 +86,7 @@ class CCLCosmology(CLMMCosmology):
         a2 = np.atleast_1d(self.get_a_from_z(z2))
         if len(a1)==1 and len(a2)!=1:
             a1 = np.repeat(a1, len(a2))
-        if len(a2)==1 and len(a1)!=1:
+        elif len(a2)==1 and len(a1)!=1:
             a2 = np.repeat(a2, len(a1))
 
         sign = np.sign(a1-a2)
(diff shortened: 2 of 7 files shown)


m-aguena:
@hsinfan1996 I added some overhead operations in the parent class to `_eval_da_z1z2` and `_eval_sigma_crit` to compute only for z1>z2 and add bad values to the rest. In this implementation, we just need to rename `_eval_da_z1z2` -> `_eval_da_z1z2_core` and `_eval_sigma_crit -> `_eval_sigma_crit_core` and remove the old verifications for this in the different backend classes.

hsinfan1996:
> @hsinfan1996 I added some overhead operations in the parent class to `_eval_da_z1z2` and `_eval_sigma_crit` to compute only for z1>z2 and add bad values to the rest. In this implementation, we just need to rename `_eval_da_z1z2` -> `_eval_da_z1z2_core` and `_eval_sigma_crit -> `_eval_sigma_crit_core` and remove the old verifications for this in the different backend classes.

Thank you @m-aguena. I renamed `_eval_da_z1z2` and `_eval_sigma_crit` as instructed, but the tests failed because
1.  `z1` or `z2` cannot be a list, though I can see the docstrings explicitly write that the input redshifts should be either an array or a float. However, in `test_theory.py` line 559, z_source is a list, and it cannot be indexed by a boolean array, so the tests failed.
2. even if `z1` and `z2` are arrays, when `z_good` is all False, `z1` and `z2` and empty arrays. They will not pass `input_validation` because `arg` is empty and there is nothing to retrieve from `arg[0]`.

I fixed the problems in my own way, but I do not think it is elegant. Feel free to make changes again to make it less complicated.

m-aguena (approved):
It looks good to me now. I would appreciate if @combet could have a final look on this PR before merging.

combet (commented):
Thanks @hsinfan1996 and @m-aguena. Before approving, 
- [ ] may I suggest to show and explain the behavior of the `eval_da_z1z2` function in the `demo_theory_functionality{_oo}.ipynb` notebooks.
- [x]  need to update the docstring of `eval_da_z1z2` (and `eval_da_a1a2`?) to say that arguments can be float or array-like

Issue #508, opened 2022-05-05
test_utils.py Line 400


hsinfan1996:
valid_type should be `float` instead of `'float'`. I know a None argument is not checked if `none_ok=True`.

PR #511, opened 2022-05-12
Issue/455/add warnings

hsinfan1996:


hsinfan1996:
Does anyone also encounter `AttributeError: 'NumCosmoCosmology' object has no attribute 'smd'` when calling `eval_sigma_crit` with NumCosmo backend?

hsinfan1996:
@m-aguena @marina-ricci Can I fix issue #508 in this PR? I just need to delete a pair of quotation marks, so I do not want to create a branch for it.

m-aguena:
> @m-aguena @marina-ricci Can I fix issue #508 in this PR? I just need to delete a pair of quotation marks, so I do not want to create a branch for it.

sure, go for it

hsinfan1996:
I added a flag `show_warning=True` to `eval_critical_surface_density` and `eval_sigma_crit` to suppress the extra warnings. Not sure if there is better way to deal with that.

hsinfan1996:
Merged to PR #512

PR #512, opened 2022-05-27
Issue/496/WL function warnings and CCL v2.6

eduardojsbarroso:


clmm/theory/ccl.py
```diff
@@ -152,5 +152,52 @@ def _eval_excess_surface_density(self, r_proj, z_cl):
         else:
             return self.eval_mean_surface_density(r_proj, z_cl) - self.eval_surface_density(r_proj, z_cl)
 
-
+    def _prepare_ccl(self, r_proj, z_cl, z_src):
```
eduardojsbarroso, line 155: I created this function to prepare the objects for the ccl functions. Since we want to be able to give a list of projected radius and source redshifts, this object turns the given data into lists, even if there is only one data for each variable.

Also, the ccl uses the scale factor instead of the redshift, so I had to compute the variables. Lastly, I use np.vectorize so this functions works the same way as the NumCosmo backend.

clmm/theory/ccl.py
```diff
@@ -152,5 +152,52 @@ def _eval_excess_surface_density(self, r_proj, z_cl):
...
+        a_cl = self.cosmo.get_a_from_z(z_cl)
+        ra_proj = [radius/a_cl for radius in r_proj_l]
+        return [ra_proj, a_cl, a_src]
+    
+    def _eval_convergence(self, r_proj, z_cl, z_src):
+        """eval convergence"""
+        ra_proj, a_cl, a_src =self._prepare_ccl(r_proj, z_cl, z_src)        
+        func = lambda ra_proj, a_cl, a_src: self.hdpm.convergence(self.cosmo.be_cosmo, ra_proj, self.mdelta, a_cl, a_src, self.mdef)
+        
+        return np.vectorize(func)(ra_proj, a_cl, a_src)
```
eduardojsbarroso, line 177: Lastly, I use np.vectorize so these functions work the same way as the NumCosmo backend.

examples/comparison_ccl_nc_theory_backend.ipynb
```diff
@@ -0,0 +1,232 @@
...
+   "cell_type": "code",
+   "execution_count": 7,
+   "id": "4dbd1763-832e-4a67-a496-140f6b315e49",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#del os.environ['CLMM_MODELING_BACKEND']\n",
+    "\n",
+    "#os.environ['CLMM_MODELING_BACKEND'] = 'ccl' \n",
+    "\n"
```
eduardojsbarroso, line 115: I was trying to use both backends to compare the results from NumCosmo and CCL, but I can't find a way to use both in the same notebook. So I ran the notebook for numcosmo, wrote down the results in the last cell, and then changed the backend to ccl.

examples/comparison_ccl_nc_theory_backend.ipynb
```diff
@@ -0,0 +1,232 @@
...
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "1.0000145112041072 7.253514366217001e-06 6.338309030485296e-05 6.338355005834388e-05\n"
+     ]
+    }
+   ],
+   "source": [
+    "#NumCosmo results\n",
+    "print(1.0000145112041072 ,7.253514366217001e-06 ,6.338309030485296e-05, 6.338355005834388e-05)"
```
eduardojsbarroso, line 207: The results seem to mach up to 5 decimals, but in the 6th decimal, ccl and numcosmo are different for some values.

eduardojsbarroso:
@pennalima Could you review this PR to see if this solves the issue? Let me know if there is something that I need to change. Also, if someone knows how I can use 2 backends in the same notebook, I would appreciate the help. @combet @m-aguena @marina-ricci .

combet:
Thanks @eduardojsbarroso - I don't think there is a way to use two backends in the same NB, but others should confirm.

eduardojsbarroso:
> Thanks @eduardojsbarroso - I don't think there is a way to use two backends in the same NB, but others should confirm. Also, the tests are currently failing. I believe this is due to spurious `if type(z_src) == float:` conditions. Could you please check it out?

Sure.  I was searching for the problem last night and I maybe know the answer, but I will investigate. It seems to be because I'm passing list variables but passing a float z_cl.

hsinfan1996:
> > We face the same problem here as in PR #510. CCL requires a_cl > a_src.
> 
> @hsinfan1996 Thanks Hsin. I have a question. Isn't that always the case when we say that the light from a source galaxy suffers lensing from a cluster? Because if the source galaxy is "behind", it has `z_src > z_cl`, and this will result in `a_cl > a_src`. Or am I getting something wrong?
> 
> And also, should we change the tests values for values that satisfy `a_cl > a_src` at least for the CCL backend?

Sorry, by that I mean CCL will output errors when a_cl<a_src. You are totally right. It is just we are trying to deal with the case where a_cl<a_src. The issue is addressed in PR #510, and I think the solution can be applied here.

eduardojsbarroso:
@combet @m-aguena @m-aguena @hsinfan1996 

Hi all. Today I found some problems in the CCL backend functions and I wanted to know if you guys have any suggestions. When I compare the magnification, magnification bias and the reduced tangential shear computed using CCL to the results using NumCosmo, there is an error of 1% for projected radii of the order of ````10^-2````. I`m attaching the pictures bellow.

As I see, the problem is comming from the fact that we are getting negative results for the magnification in this range of radii. Because of that, I get the error message saying the this leads to nan values of the magnification bias, but when I print these results, they are negative but not nan.

I don't know if there is some value that I should not be using or if this is an error from the CCL backend, since the NumCosmo results are just fine.

magnification
error_message
.

hsinfan1996:
We probably need to optimize the WL functions. At CCL level, the 4 functions only take a_len and a_source as floats, so we need to vectorize the functions on our end (I guess _prepare_ccl is added for this reason). Vectorizing makes the function ~400 times slower, but the underlying calculation is essentially the same as the original implementation.

eduardojsbarroso:
@hsinfan1996  Do you have any suggestions on how to improve the code? I did a function that prepares the CLL functions to vectorize, but I'm not sure if it is the fastest way.

hsinfan1996:
> @hsinfan1996 Do you have any suggestions on how to improve the code? I did a function that prepares the CLL functions to vectorize, but I'm not sure if it is the fastest way.

I will try something on the CCL side since in CCL's documentation a_len and a_source can be array_like.

eduardojsbarroso:
> I guess we should probably fix CCL's documentation first and then fix the functions, because in, e.g. `convergence` in `pyccl/halos/profiles.py`, `self.projected` only takes `a_lens` as a float (`a_source` can be a float or array_like). But in the next step, if `a_source` is array_like, then `a_lens` has to be array_like as well. So in the end, `a_lens` has to be a float and cannot be array_like.

That is probabily why I keep getting the error on the input scale factor, beacuse it can not be an array. Let us wait to see if they fix it after we open an issue there.

hsinfan1996:
I probably need someone to help me with https://github.com/LSSTDESC/CCL/pull/943.

eduardojsbarroso:
> I probably need someone to help me with [LSSTDESC/CCL#943](https://github.com/LSSTDESC/CCL/pull/943).

I can help. Do you want me to focus on something in particular?

hsinfan1996:
@eduardojsbarroso Thank you. I am not sure what I said in the CCL repo about our need to make array_like `a_source` supported was correct. I am afraid that I did not express our needs clearly, so it would be helpful to talk about how we use the functions in CLMM (in many cases, `a_source` has the same dimension as `r`.)
I thought it would be an easy fix and it turned out they want more than that. I think my code fulfills the minimum requirement of `a_source ` being array_like stated in the documentation with `a_source ` having the same shape as `r`. Now they want to allow `r` and `a_source` to have different shapes, which can be useful but is not required for us.

hsinfan1996:
If CCL does fix this problem in the next release, our calculations relying on CCL will need to require CCL version higher than v2.4.0. Otherwise, we will need to use our original implementation, because vectorizing the functions is way too slow. In test_mock where 260000 galaxies are generated, it takes too long to complete.

eduardojsbarroso:
> @eduardojsbarroso Thank you. I am not sure what I said in the CCL repo about our need to make array_like `a_source` supported was correct. I am afraid that I did not express our needs clearly, so it would be helpful to talk about how we use the functions in CLMM (in many cases, `a_source` has the same dimension as `r`.) I thought it would be an easy fix and it turned out they want more than that. I think my code fulfills the minimum requirement of `a_source ` being array_like stated in the documentation with `a_source ` having the same shape as `r`. Now they want to allow `r` and `a_source` to have different shapes, which can be useful but is not required for us.

I wrote a comment on the CCL issue trying to explain our case. Please correct me if I understood something wrong. But I feel like the code you edited is good enough for us, and I don't really see how allowing `r` and `a_source` to have different shapes will be usefull, but this would represent a general case that would fill our needs.

eduardojsbarroso:
> If CCL does fix this problem in the next release, our calculations relying on CCL will need to require CCL version higher than v2.4.0. Otherwise, we will need to use our original implementation, because vectorizing the functions is way too slow. In test_mock where 260000 galaxies are generated, it takes more than 30 sec to complete.

I think that if they don't implement this feature, we can think of a way to upgrade the code that I made to prepare the functions for the vectorization. Does the NumCosmo backend takes this long aswell? Because we use `np.vectorize` there also, so  if not, I think that the '_prepare' function that I created is maybe generating this lag.

hsinfan1996:
@eduardojsbarroso Thank you. I did some tests on NC backend, it is two times slower than the original implementation for CCL backend, but ~100 times faster than vectorizing CCL function.
image

hsinfan1996:
The four functions can now take array_like `a_source` after https://github.com/LSSTDESC/CCL/pull/943 is merged. If we implement the functions, only CCL > v2.4.1  will work properly.
Do we use our current calculations for the older versions of CCL?
We still need to make sure only `a_source` less `a_lens` is passed to the functions.

hsinfan1996:
@eduardojsbarroso I implemented the updated CCL functions following the logic in PR #510. I am sorry for not reusing most of your code, but at the same time, I will have to discard a lot of changes in PR #511 since the two PRs are somewhat related.

eduardojsbarroso:
> @eduardojsbarroso I implemented the updated CCL functions following the logic in PR #510. I am sorry for not reusing most of your code, but at the same time, I will have to discard a lot of changes in PR #511 since the two PRs are somewhat related.

No problem. Thanks for the help.

eduardojsbarroso:
@hsinfan1996  I just reviewed the code and everything looks good to me.  The only thing that I thought off is that the tests in `test_theory.py` uses functions with `z_source < z_cl` and then they just print out zeros when we test it, but shouldn't we maybe implement some other tests with non-zero results?

 I think that NumCosmo and CCL have different outputs and maybe we should see if the difference between them respect the tolerance error. But maybe this should be a new issue.

hsinfan1996:
> @hsinfan1996 I just reviewed the code and everything looks good to me. The only thing that I thought off is that the tests in `test_theory.py` uses functions with `z_source < z_cl` and then they just print out zeros when we test it, but shouldn't we maybe implement some other tests with non-zero results?
> 
> I think that NumCosmo and CCL have different outputs and maybe we should see if the difference between them respect the tolerance error. But maybe this should be a new issue.

@eduardojsbarroso Thank you. Maybe we can add tests for an array of radius and an array of z_cource with mixed `z_source < z_cl` and `z_source > z_cl`.

clmm/theory/func_layer.py
```diff
@@ -595,42 +603,51 @@ def compute_reduced_tangential_shear(
             * `virial` - not in cluster_toolkit;
 
     alpha_ein : float, optional
-        If `halo_profile_model=='einasto'`, set the value of the Einasto slope. Option only available
-        for the NumCosmo backend
+        If `halo_profile_model=='einasto'`, set the value of the Einasto slope.
+        Option only availablefor the NumCosmo backend
```
combet, line 607: availabefor --> space missing

combet:
@hsinfan1996, @eduardojsbarroso  - I started looking at this PR. I see a lot of changed mostly due to docstrings formatting. Could you please summarize the main things that where done, and where, so I can first focus on the new implementation. Thanks!

hsinfan1996:
1. `cosmology.ccl.py`: `_eval_sigma_crit_core` now uses CCL function. Old code are commented out.
2. `self.smd = Nc.WLSurfaceMassDensity.new(self.dist)` and `self.smd.prepare_if_needed(self.be_cosmo)` are moved from `theory.numcosmo.py` to `cosmology.numcosmo.py` to fix the issue where the method `eval_sigma_crit` of a NumCosmoCosmology will fail.
3. Add `show_warning` flag to `eval_sigma_crit` to suppress extra warnings.
4. `_eval_convergence_core`, `_eval_tangential_shear_core`, `_eval_reduced_tangential_shear_sp_core` and `_eval_magnification_core` use CCL functions if CCL>=2.5.0. Otherwise, old calculations will be used and there will warnings about using old CCL.
5. Add `show_warning` flag to `eval_critical_surface_density` to suppress extra warnings. 
6. `theory.parent_class.py`: `_eval_tangential_shear`, `_eval_convergence`, `_eval_reduced_tangential_shear_sp` and `_eval_magnification` are renmed to `*_core`.
7. `utils.py`: There are changes to `compute_for_good_redshifts` in order to reuse it for `eval_tangential_shear`, `eval_convergence`, `eval_reduced_tangential_shear` when `z_src_model == 'single_plane' `and `eval_magnification`
8. `test_cosmo_parent.py`: Add a test on raising UserWarning for `cosmo.eval_sigma_crit`
9. `test_theory_parent.py`: Add tests on raising warnings when z_src<z_lens and when using CCL<2.5.0

hsinfan1996:
I have not tested the results given by `eval_sigma_crit` before and after switching to the CCL function. By looking at the calculation, there should not be any difference. Even if they are different, we cannot do anything about it other than warning the users.

clmm/cosmology/ccl.py
```diff
@@ -104,15 +103,23 @@ def _eval_da_z1z2_core(self, z1, z2):
...
-        return (cte*Ds/(Dl*Dls))*self.cor_factor
+        #return (cte*Ds/(Dl*Dls))*self.cor_factor
+
+        # Use CCL function
+        a_len = self.get_a_from_z(z_len)
+        a_src = self.get_a_from_z(z_src)
+        if np.iterable(a_src):
+            a_len = np.full_like(a_src, a_len)
+
+        return self.be_cosmo.sigma_critical(a_len, a_src)
```
hsinfan1996, line 122: 1

clmm/cosmology/numcosmo.py
```diff
@@ -39,6 +39,8 @@ def __init__(self, dist=None, dist_zmax=15.0, **kwargs):
             self.set_dist(dist)
         else:
             self.set_dist(Nc.Distance.new(dist_zmax))
+        self.smd = Nc.WLSurfaceMassDensity.new(self.dist)
+        self.smd.prepare_if_needed(self.be_cosmo)
```
hsinfan1996, line 43: 2

clmm/cosmology/parent_class.py
```diff
@@ -401,7 +403,7 @@ def mpc2rad(self, dist1, redshift):
             validate_argument(locals(), 'redshift', float, argmin=0, eqmin=True)
         return dist1/self.eval_da(redshift)
 
-    def eval_sigma_crit(self, z_len, z_src):
+    def eval_sigma_crit(self, z_len, z_src, show_warning=True):
```
hsinfan1996, line 406: 3

clmm/cosmology/parent_class.py
```diff
@@ -423,13 +425,15 @@ def eval_sigma_crit(self, z_len, z_src):
...
-    def _eval_sigma_crit(self, z_len, z_src):
+    def _eval_sigma_crit(self, z_len, z_src, show_warning=False):
+        warning_msg = '\nSome source redshifts are lower than the cluster redshift.'+\
+        '\nSigma_crit = np.inf for those galaxies.'
         return compute_for_good_redshifts(
             self._eval_sigma_crit_core, z_len, z_src, np.inf,
-            error_message='Some source redshifts are lower than the cluster redshift. '
-                          'Returning Sigma_crit = np.inf for those galaxies.')
+            z1_arg_name='z_len', z2_arg_name='z_src',
+            warning_message=warning_msg, show_warning=show_warning)
```
hsinfan1996, line 436: 3

clmm/theory/ccl.py
```diff
@@ -156,13 +157,67 @@ def _eval_excess_surface_density(self, r_proj, z_cl):
...
+            warnings.warn('\nOlder version of CCL detected')
+            return super()._eval_magnification_core(r_proj, z_cl, z_src)
+
+        a_cl = self.cosmo.get_a_from_z(z_cl)
+        a_src = self.cosmo.get_a_from_z(z_src)
+
+        func = self.hdpm.magnification
+        res = func(self.cosmo.be_cosmo, r_proj/a_cl, self.mdelta, a_cl, a_src, self.mdef)
+
+        return res
```
hsinfan1996, line 220: 4

clmm/theory/parent_class.py
```diff
@@ -165,8 +169,8 @@ def _eval_excess_surface_density(self, r_proj, z_cl):
     # 3. Functions that can be used by all subclasses
 
 
-    def _eval_critical_surface_density(self, z_len, z_src):
-        return self.cosmo.eval_sigma_crit(z_len, z_src)
+    def _eval_critical_surface_density(self, z_len, z_src, show_warning=False):
+        return self.cosmo.eval_sigma_crit(z_len, z_src, show_warning=show_warning)
```
hsinfan1996, line 173: 5

clmm/theory/parent_class.py
```diff
@@ -210,22 +214,22 @@ def __integrand__( l , theta ):
...
 
-    def _eval_reduced_tangential_shear_sp(self, r_proj, z_cl, z_src):
+    def _eval_reduced_tangential_shear_sp_core(self, r_proj, z_cl, z_src):
         kappa = self.eval_convergence(r_proj, z_cl, z_src)
         gamma_t = self.eval_tangential_shear(r_proj, z_cl, z_src)
         return compute_reduced_shear_from_convergence(gamma_t, kappa)
 
-    def _eval_magnification(self, r_proj, z_cl, z_src):
+    def _eval_magnification_core(self, r_proj, z_cl, z_src):
         kappa = self.eval_convergence(r_proj, z_cl, z_src)
```
hsinfan1996, line 233: 6

clmm/theory/parent_class.py
```diff
@@ -388,23 +393,27 @@ def eval_3d_density(self, r3d, z_cl, verbose=False):
 
         return self._eval_3d_density(r3d=r3d, z_cl=z_cl)
 
-    def eval_critical_surface_density(self, z_len, z_src=None, use_pdz=False, pzbins=None, pzpdf=None):
+    def eval_critical_surface_density(self, z_len, z_src=None,
+                                      use_pdz=False, pzbins=None, pzpdf=None,
+                                      show_warning=True):
```
hsinfan1996, line 398: 5

clmm/theory/parent_class.py
```diff
@@ -428,19 +438,22 @@ def eval_critical_surface_density(self, z_len, z_src=None, use_pdz=False, pzbins
...
+        """
 
         if self.validate_input:
             validate_argument(locals(), 'z_len', float, argmin=0)
             validate_argument(locals(), 'z_src', 'float_array', argmin=0, none_ok=True)
 
         if use_pdz is False:
-            return self._eval_critical_surface_density(z_len=z_len, z_src=z_src)
+            return self._eval_critical_surface_density(z_len=z_len, z_src=z_src,
+                                                       show_warning=show_warning)
```
hsinfan1996, line 452: 5

clmm/theory/parent_class.py
```diff
@@ -598,18 +612,21 @@ def eval_tangential_shear(self, r_proj, z_cl, z_src, verbose=False):
...
-
         if self.halo_profile_model=='einasto' and verbose:
             print(f"Einasto alpha = {self._get_einasto_alpha(z_cl=z_cl)}")
 
-        return self._eval_tangential_shear(r_proj=r_proj, z_cl=z_cl, z_src=z_src)
+        warning_msg = '\nSome source redshifts are lower than the cluster redshift.'+\
+        '\nShear = 0 for those galaxies.'
+        return compute_for_good_redshifts(self._eval_tangential_shear_core,
+                                          z_cl, z_src, 0., warning_msg,
+                                          'z_cl', 'z_src', r_proj)
```
hsinfan1996, line 629: 7

clmm/theory/parent_class.py
```diff
@@ -633,22 +650,29 @@ def eval_convergence(self, r_proj, z_cl, z_src, verbose=False):
...
-
         if self.halo_profile_model=='einasto' and verbose:
             print(f"Einasto alpha = {self._get_einasto_alpha(z_cl=z_cl)}")
 
-        return self._eval_convergence(r_proj=r_proj, z_cl=z_cl, z_src=z_src)
+        warning_msg = '\nSome source redshifts are lower than the cluster redshift.'+\
+        '\nConvergence = 0 for those galaxies.'
+        return compute_for_good_redshifts(self._eval_convergence_core,
+                                          z_cl, z_src, 0., warning_msg,
+                                          'z_cl', 'z_src', r_proj)
```
hsinfan1996, line 667: 7

clmm/theory/parent_class.py
```diff
@@ -698,27 +733,36 @@ def eval_reduced_tangential_shear(self, r_proj, z_cl, z_src, z_src_model='single
             print(f"Einasto alpha = {self._get_einasto_alpha(z_cl=z_cl)}")
 
         if z_src_model == 'single_plane':
-            gt = self._eval_reduced_tangential_shear_sp(r_proj, z_cl, z_src)
+            warning_msg = '\nSome source redshifts are lower than the cluster redshift.'+\
+            '\nReduced_shear = 0 for those galaxies.'
+            gt = compute_for_good_redshifts(self._eval_reduced_tangential_shear_sp_core,
+                                            z_cl, z_src, 0., warning_msg,
+                                            'z_cl', 'z_src', r_proj)
```
hsinfan1996, line 740: 7

clmm/theory/parent_class.py
```diff
@@ -754,11 +798,14 @@ def eval_magnification(self, r_proj, z_cl, z_src, verbose=False):
...
-
         if self.halo_profile_model=='einasto' and verbose:
             print(f"Einasto alpha = {self._get_einasto_alpha(z_cl=z_cl)}")
 
-        return self._eval_magnification(r_proj=r_proj, z_cl=z_cl, z_src=z_src)
+        warning_msg = '\nSome source redshifts are lower than the cluster redshift.'+\
+        '\nMagnification = 1 for those galaxies.'
+        return compute_for_good_redshifts(self._eval_magnification_core,
+                                          z_cl, z_src, 1., warning_msg,
+                                          'z_cl', 'z_src', r_proj)
```
hsinfan1996, line 808: 7

clmm/utils.py
```diff
@@ -607,8 +613,10 @@ def _integ_pzfuncs(pzpdf, pzbins, zmin, kernel=lambda z: 1., ngrid=1000):
     kernel_matrix = kernel(z_grid)
     return scipy.integrate.simps(pz_matrix*kernel_matrix, x=z_grid, axis=1)
 
-def compute_for_good_redshifts(function, z1, z2, bad_value, error_message):
-    """Computes function only for z1>z2, the rest is filled with bad_value
+def compute_for_good_redshifts(function, z1, z2, bad_value, warning_message,
+                               z1_arg_name='z1', z2_arg_name='z2', r_proj=None,
+                               show_warning=True):
+    """Computes function only for `z1` < `z2`, the rest is filled with bad_value
```
hsinfan1996, line 619: 7

clmm/utils.py
```diff
@@ -619,30 +627,51 @@ def compute_for_good_redshifts(function, z1, z2, bad_value, error_message):
...
-                    np.array(z2)[z_good] if np.iterable(z2) else z2)
+                kwargs[z1_arg_name] = np.array(z1)[z_good] if np.iterable(z1) else z1
+                kwargs[z2_arg_name] = np.array(z2)[z_good] if np.iterable(z2) else z2
+                res[z_good] = function(**kwargs)
         else:
             res = bad_value
     else:
-        res = function(z1, z2)
+        res = function(**kwargs)
     return res
```
hsinfan1996, line 668: 7

tests/test_cosmo_parent.py
```diff
@@ -115,6 +114,8 @@ def test_cosmo_basic(modeling_data, cosmo_init):
     """ Unit tests abstract class z and a methdods """
     reltol = modeling_data['cosmo_reltol']
     cosmo = theo.Cosmology(**cosmo_init)
+    # Test warnings
+    assert_warns(UserWarning, cosmo.eval_sigma_crit, 0.3, [0.2, 0.3, 0.4])
```
hsinfan1996, line 118: 8

tests/test_theory_parent.py
```diff
@@ -111,38 +118,101 @@ def test_instantiate(modeling_data):
...
+        import pyccl as ccl
+        if not version.parse(ccl.__version__) < version.parse('2.5.0'):
+            ccl.__version__='2.4.0'
+        #test warnings from using CCL<2.5.0
+        assert_warns(UserWarning, mod.eval_convergence, [0.3], 0.3, [0.4])
+        assert_warns(UserWarning, mod.eval_tangential_shear, [0.3], 0.3, [0.4])
+        assert_warns(UserWarning, mod.eval_reduced_tangential_shear,
+                     [0.3], 0.3, [0.4])
+        assert_warns(UserWarning, mod.eval_magnification, [0.3], 0.3, [0.4])
+
```
hsinfan1996, line 215: 9

combet:
Thanks a lot @hsinfan1996 - I'll get to it as soon as possible

hsinfan1996:
> Thanks a lot @hsinfan1996 - I'll get to it as soon as possible

No hurry. Take your time.

hsinfan1996:
I would like to wait for the next CCL release again in order to get the version number (2.5.3 probably), so when we are implementing new functionalities (in different PRs), we don't need to check for many different version numbers of CCL.

m-aguena:
[from 25/11/2022 tag-up]: still waiting for CCL release

hsinfan1996:
Waiting for PR #557.

hsinfan1996:
> Any idea why this check is failing? I am seeing the same failure on another pr.

PR #576 will fix the problem.

clmm/cosmology/ccl.py
```diff
@@ -104,15 +103,21 @@ def _eval_da_z1z2_core(self, z1, z2):
...
+        #cte = ccl.physical_constants.CLIGHT**2 / \
+        #    (4.0*np.pi*ccl.physical_constants.GNEWT *
+        #     ccl.physical_constants.SOLAR_MASS)*ccl.physical_constants.MPC_TO_METER
 
-        Ds = self._eval_da_z1z2_core(0, z_src)
-        Dl = self._eval_da_z1z2_core(0, z_len)
-        Dls = self._eval_da_z1z2_core(z_len, z_src)
+        #Ds = self._eval_da_z1z2_core(0, z_src)
+        #Dl = self._eval_da_z1z2_core(0, z_len)
+        #Dls = self._eval_da_z1z2_core(z_len, z_src)
```
combet, line 112: Could you delete those commented-out lines?

clmm/theory/ccl.py
```diff
@@ -130,51 +132,74 @@ def _get_einasto_alpha(self, z_cl=None):
...
+    def _eval_convergence_core(self, r_proj, z_cl, z_src):
+        """eval convergence"""
+
+        a_cl = self.cosmo.get_a_from_z(z_cl)
+        a_src = self.cosmo.get_a_from_z(z_src)
+
+        func = self.hdpm.convergence
+        res = func(self.cosmo.be_cosmo, r_proj/a_cl, self.mdelta, a_cl, a_src, self.mdef)
+
+        return res
```
combet, line 170: Maybe compact those 3 lines in 1?

`return self.hdpm.convergence(self.cosmo.be_cosmo, r_proj/a_cl, self.mdelta, a_cl, a_src, self.mdef)`

clmm/theory/ccl.py
```diff
@@ -130,51 +132,74 @@ def _get_einasto_alpha(self, z_cl=None):
...
+
+    def _eval_tangential_shear_core(self, r_proj, z_cl, z_src):
+
+        a_cl = self.cosmo.get_a_from_z(z_cl)
+        a_src = self.cosmo.get_a_from_z(z_src)
+
+        func = self.hdpm.shear
+        res = func(self.cosmo.be_cosmo, r_proj/a_cl, self.mdelta, a_cl, a_src, self.mdef)
+
+        return res
```
combet, line 180: If agreed above for `_eval_convergence_core`, do the same here.

clmm/theory/ccl.py
```diff
@@ -130,51 +132,74 @@ def _get_einasto_alpha(self, z_cl=None):
...
+    def _eval_reduced_tangential_shear_core(self, r_proj, z_cl, z_src):
+        """eval reduced tangential shear with all background sources at the same plane"""
+
+        a_cl = self.cosmo.get_a_from_z(z_cl)
+        a_src = self.cosmo.get_a_from_z(z_src)
+
+        func = self.hdpm.reduced_shear
+        res = func(self.cosmo.be_cosmo, r_proj/a_cl, self.mdelta, a_cl, a_src, self.mdef)
+
+        return res
```
combet, line 191: If agreed above for `_eval_convergence_core`, do the same here.

clmm/theory/ccl.py
```diff
@@ -130,51 +132,74 @@ def _get_einasto_alpha(self, z_cl=None):
...
+    def _eval_magnification_core(self, r_proj, z_cl, z_src):
+        """eval magnification"""
+
+        a_cl = self.cosmo.get_a_from_z(z_cl)
+        a_src = self.cosmo.get_a_from_z(z_src)
+
+        func = self.hdpm.magnification
+        res = func(self.cosmo.be_cosmo, r_proj/a_cl, self.mdelta, a_cl, a_src, self.mdef)
+
+        return res
```
combet, line 202: If agreed above for `_eval_convergence_core`, do the same here.

combet (changes requested):
Thank you - I just have a few minor comments. All looks good otherwise.

hsinfan1996:
@combet Thank you. I made changes following the comments.

clmm/theory/ccl.py
```diff
@@ -130,51 +132,64 @@ def _get_einasto_alpha(self, z_cl=None):
...
-        else:
-            return self.eval_mean_surface_density(r_proj, z_cl) \
-                    - self.eval_surface_density(r_proj, z_cl)
 
+        return (self.hdpm.cumul2d(self.cosmo.be_cosmo, r_proj/a_cl, self.__mdelta_cor,
+                                  a_cl, self.mdef) -
+                    self.hdpm.projected(self.cosmo.be_cosmo, r_proj/a_cl, self.__mdelta_cor,
+                                        a_cl, self.mdef))*self.cor_factor/a_cl**2
+
+    def _eval_convergence_core(self, r_proj, z_cl, z_src):
```
m-aguena, line 161: @hsinfan1996 shouldn't these functions have `self.__mdelta_cor` instead of `self.mdelta`  as input? I guess the correction factor is small enough for the tests to pass, but this should be checked.

hsinfan1996: Fixed.

m-aguena:
Thanks @hsinfan1996 , everything is good now.
I just decided to simplify the code a bit, so I will ask you to check on my latest changes (pinging @combet too).

hsinfan1996 (approved):

combet (approved):
All good! Thank you

Issue #509, opened 2022-05-05
Allowing different type of `z_src`


marina-ricci:
In PR #467 we decided it would be good to allow different option for the parameter `z_src`, based on `z_src_model` : float, array or function.

PR #517, opened 2022-06-20
Issue_509_allow_different_zsrc_types

marina-ricci:
This PR aims to address issue #509.  To dos : 

- [x] change "single plane" option name to "discrete", as it allows z_src to be an array and is thus misleading. 
- [x] allow `z_source` to be a function (in place of `z_distrib_func`) if ` z_src_model` uses a function.
- [x] generalise this to other computation than reduced tengential shear   (` z_src_model` should be either discrete or a function, and for reduced tengential shear we need an extra keyword for the  specific developement).
- [x] add “Not supported” for magnification bias if it makes sense
- [x] update the tests
- [x] write new tests
- [x] check if notebooks run 
- [x] make a new notebook to show the functionalities.


marina-ricci:
@m-aguena this is what I came up with for the `compute_reduced_tangential_shear` function.

clmm/theory/func_layer.py
```diff
@@ -577,22 +577,26 @@ def compute_reduced_tangential_shear(
...
         Galaxy cluster redshift
     z_source : array_like, float, one-parameter function or dict
-        Source redshift information depending on the `z_src_model` parameter. The options available are:       
-            * Background source galaxy redshift(s) (array_like, float) if `z_src_model`='discrete' (default),
-            * Background source galaxy redshift distribution function (one-parameter function) if `z_src_model`=`distribution`,
-            * Lensing efficiency parameter(s) averaged over the galaxy redshift distribution (dict) if `z_src_model`=`averaged lensing efficiency`.
-            The dictionnary should contain a 'beta_s_mean' key and a 'beta_s_square_mean' key with the appropriate corresponding values :
-                 *beta_s_mean: array_like, float
-                    Lensing efficiency averaged over the galaxy redshift distribution. 
+        Source redshift information depending on the `z_src_model` parameter:
```
marina-ricci, line 580: Thank you @m-aguena for editing the docstrings. For `z_source` description I think it is better to not start by the`z_src_model` value as it is a bit misleading.

clmm/theory/func_layer.py
```diff
@@ -619,22 +623,27 @@ def compute_reduced_tangential_shear(
...
-        Expression for the compupuation of the approximated averaged reduced tangential shear. This is not used if `z_src_model`='discrete'. The supported options are:
-            
-            * `applegate14` (default): use the equation (6) in Weighing the Giants - III (Applegate et al. 2014; https://arxiv.org/abs/1208.0605).
-            * `schrabback18`: use the equation (12) in Cluster Mass Calibration at High Redshift (Schrabback et al. 2017; https://arxiv.org/abs/1611.03866).              
-            
+            or known individual source galaxy redshifts (if `z_source` is an array and \
+            `r_proj` is a float);
+            * `distribution` : sources follow a redshift distribution function. In this case \
+            the averaged lensing efficiency parameter(s) will be computed and used to \
+            approximate the reduced tangential shear;
```
hsinfan1996, line 630: For the documentation to compile, we will need to omit the backslashes and align the beginning of the next line with that (actual start of the text, not the * symbol and space) of the previous one by adding two spaces. Same for the other items in the lists. Also check out here for the [sphinx documentation](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#lists-and-quote-like-blocks).

clmm/theory/func_layer.py
```diff
@@ -619,22 +623,27 @@ def compute_reduced_tangential_shear(
...
+            * `averaged lensing efficiency` : the source redshift distribution is described by \
+            the averaged lensing efficiency parameter(s). In this case these parameters will \
+            directly be used to approximate the reduced tangential shear;
+
+    gt_equation: str, optional
+        Expression for the compupuation of the approximated averaged reduced tangential shear. \
+        This is not used if `z_src_model`='discrete'. The supported options are:
```
hsinfan1996, line 637: Need to be `` `z_src_model` = 'discrete'`` instead of  `` `z_src_model`='discrete'`` for the documentation to compile.

m-aguena:
@marina-ricci I wrote another possibility of having this implementation in the parent class, where we have 3 arguments describing all cases we want:

- `z_src_info` : array_like, float, function
            Information on the background source galaxy redshift(s). Value required depends on z_src_type (see below).
- `z_src_type` : str, optional
            Type of redshift information provided, it describes z_src_info.
            The following supported options are:
  - `discrete` (default) : The redshift of sources is provided by `z_src_info`.
                  It can be individual redshifts for each source galaxy when `z_source` is an array
                  or all sources are at the same redshift when `z_source` is a float.
  - `distribution` : A redshift distribution function is provided by `z_src_info`.
                  `z_scr_info` must be a one dimentional function. If `z_scr_info=None`,
                  the Chang et al (2013) distribution function is used.
  - `beta` : The averaged lensing efficiency is provided by `z_src_info`.
                  `z_scr_info` must be a tuple containing (beta_s, beta_s^2).

- `model` : str, optional
            Type of computation to be made for reduced shears, options are:
  - `individual` (default): Computes reduced shear for each `r_proj, z_src_info` pair individually. It requires `z_src_type` to be `discrete`.
  - `applegate14` : `z_scr_type` must be either `beta`, or `distribution`
  - `schrabback18` : `z_scr_type` must be either `beta`, or `distribution` 


It's all in PR #536 , which is branched from this one. If we decide to use it, I can merge it back to this branch.

marina-ricci:
Hi @m-aguena ! Thanks. That's not another possibility, it is more or less exactly what I had in mind and was working on.

clmm/theory/parent_class.py
```diff
@@ -692,36 +712,57 @@ def eval_reduced_tangential_shear(self, r_proj, z_cl, z_src, z_src_model='single
         if self.validate_input:
             validate_argument(locals(), 'r_proj', 'float_array', argmin=0)
             validate_argument(locals(), 'z_cl', float, argmin=0)
-            validate_argument(locals(), 'z_src', 'float_array', argmin=0)
+            validate_argument(locals(), 'z_src_info', str)
+            validate_argument(locals(), 'approx', str, none_ok=True)
+            if z_src_info=='discrete':
+                validate_argument(locals(), 'z_src_info', 'float_array', argmin=0)
```
hsinfan1996, line 718: Related to the comment on the docstrings, `'z_src_info'` would be `'z_src'`. The same applies to the next two `validate_argument`.

clmm/theory/func_layer.py
```diff
@@ -560,8 +562,8 @@ def compute_convergence(r_proj, mdelta, cdelta, z_cluster, z_source, cosmo, delt
 def compute_reduced_tangential_shear(
         r_proj, mdelta, cdelta, z_cluster, z_source, cosmo,
         delta_mdef=200, halo_profile_model='nfw', massdef='mean', alpha_ein=None,
-        z_src_model='single_plane', beta_s_mean=None, beta_s_square_mean=None,
-        z_distrib_func=None, validate_input=True, verbose=False):
+        z_src_model='discrete', beta_s_mean=None, beta_s_square_mean=None,
+        z_distrib_func=None, gt_equation='applegate14', validate_input=True, verbose=False):
```
hsinfan1996, line 566: I know the tests are failing, but changing `gt_equation='applegate14'` to ` gt_equation=None` fixes a failed test from `test_mockdata.py`.

hsinfan1996: Ok. Not very sure why we default `gt_equation` to `'applegate14'`, which is not compatible with `z_src_model='discrete'`. The failed test should probably be fixed from `test_mockdata.py` itself.

clmm/support/mock_data.py
```diff
@@ -270,7 +270,7 @@ def _generate_galaxy_catalog(cluster_m, cluster_z, cluster_c, cosmo, ngals,
                                     cosmo=cosmo, delta_mdef=delta_so,
                                     halo_profile_model=halo_profile_model,
                                     massdef=massdef,
-                                    z_src_model='single_plane')
+                                    z_src_model='discrete')
```
hsinfan1996, line 273: `z_src_model` --> `z_src_info` fixes the errors from `test_mockdata`.

clmm/support/mock_data.py
```diff
@@ -279,7 +279,7 @@ def _generate_galaxy_catalog(cluster_m, cluster_z, cluster_c, cosmo, ngals,
                                 z_source=galaxy_catalog['ztrue'], cosmo=cosmo,
                                 delta_mdef=delta_so, halo_profile_model=halo_profile_model,
                                 massdef=massdef,
-                                z_src_model='single_plane')
+                                z_src_model='discrete')
```
hsinfan1996, line 282: `z_src_model` --> `z_src_info` fixes the errors from `test_mockdata`.

m-aguena:
@marina-ricci just so we don't get lost. This is the approach we decided on that telecon:

- `z_src` : array_like, float, function
            Information on the background source galaxy redshift(s). Value required depends on `z_src_info` (see below).
- `z_src_info` : str, optional
            Type of redshift information provided, it describes z_src.
            The following supported options are:
  - `discrete` (default) : The redshift of sources is provided by `z_src`.
                  It can be individual redshifts for each source galaxy when `z_src` is an array
                  or all sources are at the same redshift when `z_src` is a float.
  - `distribution` : A redshift distribution function is provided by `z_src`.
  - `beta` : The averaged lensing efficiency is provided by `z_src`.
                  `z_src` must be a tuple containing (beta_s, beta_s^2).

- `approx` : str, None, optional
            Type of computation to be made for reduced shears, options are:
  - `None` (default): Full computation is made for each `r_proj, z_src` pair individually. It requires `z_src_info` to be `discrete`.
  - `applegate14` : `z_scr_type` must be either `beta`, or `distribution`
  - `schrabback18` : `z_scr_type` must be either `beta`, or `distribution`

m-aguena:
One other thing we decided to do was to remove the default case of `z_src=None,  z_src_info='distribution'` from going to Chang et al (2013). So we still need to make `_chang_z_distrib` available to the user

marina-ricci:
I updated tangential shear, convergence, and magnification. I need to think more for `eval_critical_surface_density` and `magnification_bias`.

clmm/theory/parent_class.py
```diff
@@ -604,14 +628,39 @@ def eval_tangential_shear(self, r_proj, z_cl, z_src, verbose=False):
...
+            gammat_inf = self._eval_tangential_shear(r_proj=r_proj, z_cl=z_cl, z_src=z_inf)
+            if z_src_info=='beta':
+                # z_src (tuple) is (beta_s_mean, beta_s_square_mean)
+                beta_s_mean, beta_s_square_mean = z_src
+            elif z_src_info=='distribution':
+                # z_src (function) if PDZ
+                beta_s_mean = compute_beta_s_mean(z_cl, z_inf, self.cosmo, z_distrib_func=z_src)
+                beta_s_square_mean = compute_beta_s_square_mean(z_cl, z_inf, self.cosmo,
+                                                                z_distrib_func=z_src) 
+            gammat = beta_s_mean * gammat_inf
```
hsinfan1996, line 659: Need to raise ValueError when `z_src_info` is not `'discrete'`, `'distribution'` or `'beta'`. Otherwise, we will get `UnboundLocalError` from `test_theory` and will not pass line 451 of `test_theory`. Same for `eval_convergence`.

marina-ricci:
I came across some problems, after discussions we agree on : 
- only test if z_s < z_cl if z_src is discrete. A better handling of the warning and a generalization to all function will be done in PR #512.
- not `modify eval_critical_surface_density` this will be address by issue #538

m-aguena:
@marina-ricci I also added a file to have the redshift distribution functions, this would remove the necessity of having `z_distrib_model`. For now I just imported the current ones there, but if we decide to use it, we can move the distributions from utils to there.
To use them now, you just have to do `from clmm.z_distributions import chang2013, desc_srd`

clmm/theory/parent_class.py
```diff
@@ -906,3 +1094,22 @@ def convert_mass_concentration(self, z_cl, massdef=None, delta_mdef=None,
...
+
+    def _validate_z_src(self, loc_dict):
+        r"""Validation for z_src
+
+        Parameters
+        ----------
+        locals_dict: dict
+            Should be the call locals()
+        """
+        if loc_dic['z_src_info']=='discrete':
```
hsinfan1996, line 1106: `loc_dic` --> `loc_dict`. The same applies to the two elif's below and `beta_info`.

tests/test_theory_parent.py
```diff
@@ -129,18 +125,12 @@ def test_instantiate(modeling_data):
     
     beta_s_square_test = compute_beta_s_square_mean(z_cl, 1000., mod.cosmo)
     beta_s_test = compute_beta_s_mean(z_cl, 1000., mod.cosmo)   
-    reduced_shear = mod.eval_reduced_tangential_shear(r_proj, z_cl, z_src, 'applegate14', beta_s_mean, beta_s_square_mean)
+    reduced_shear = mod.eval_reduced_tangential_shear(r_proj, z_cl, None, 'distribution', 'applegate14')
```
hsinfan1996, line 128: first do `from clmm.z_distributions import chang2013, desc_srd` in the beginning of the file. Then changing `None` to `chang13` will fix the test.

tests/test_theory_parent.py
```diff
@@ -129,18 +125,12 @@ def test_instantiate(modeling_data):
...
     assert_allclose(reduced_shear, beta_s_test* shear_inf/(1.0 - beta_s_square_test / beta_s_test * convergence_inf), rtol=1.0e-12)
     
-    reduced_shear = mod.eval_reduced_tangential_shear(r_proj, z_cl, np.repeat(z_src, len(r_proj)), 'applegate14', beta_s_mean, beta_s_square_mean)
-    assert_allclose(reduced_shear, beta_s_test * shear_inf/(1.0 - beta_s_square_test / beta_s_test * convergence_inf), rtol=1.0e-12)
-    
-    reduced_shear = mod.eval_reduced_tangential_shear(r_proj, z_cl, z_src, 'schrabback18', beta_s_mean, beta_s_square_mean)
+    reduced_shear = mod.eval_reduced_tangential_shear(r_proj, z_cl, None, 'distribution', 'schrabback18')
```
hsinfan1996, line 131: Changing `None` to `chang13` will fix the test.

hsinfan1996:
I fixed the CI tests, so now we can see the real coveralls percentage.

hsinfan1996:
`mu = 1 / ((1 - beta_s_mean*kappa_inf)**2 - (beta_s_mean*gamma_inf)**2)` is better than the approximation we have right now, but it is not good enough (<1% difference).

marina-ricci:
> `mu = 1 / ((1 - beta_s_mean*kappa_inf)**2 - (beta_s_mean*gamma_inf)**2)` is better than the approximation we have right now, but it is not good enough (<1% difference).

How do you justify this approximation ?

hsinfan1996:
> > `mu = 1 / ((1 - beta_s_mean*kappa_inf)**2 - (beta_s_mean*gamma_inf)**2)` is better than the approximation we have right now, but it is not good enough (<1% difference).
> 
> @hsinfan1996 : How do you justify this approximation ?

That was my guess.
From https://arxiv.org/pdf/astro-ph/9910487.pdf
image
Tried the expressions without approximation and compared the results in the notebook.
For reduced shear, I tried the code below and got a maximum 9% difference.

```
z_inf = 1000. #np.inf # INF or a very large number
def integrand(z_i, r_proj):
    return (compute_beta_s(z_i, z_cl, z_inf, self.cosmo)\
            *self._eval_tangential_shear(r_proj, z_cl, z_src=z_inf)*z_src(z_i)\
    /(1-compute_beta_s(z_i, z_cl, z_inf, self.cosmo)\
      *self._eval_convergence(r_proj, z_cl, z_src=z_inf)))
gt = np.zeros_like(r_proj)
for i, r in enumerate(r_proj):
    gt[i] = quad(integrand, z_cl, 1000, r)[0] / quad(z_src, z_cl, 1000)[0]
```

For magnification, I tried the code below and got a maximum 2% difference. (Worse than the not-so-wild guess I proposed)

```
z_inf = 1000. #np.inf # INF or a very large number
def integrand(z_i, r_proj):
    return z_src(z_i)/((1-compute_beta_s(z_i, z_cl, z_inf, self.cosmo)\
                        *self._eval_convergence(r_proj, z_cl, z_src=z_inf))**2\
                       -(compute_beta_s(z_i, z_cl, z_inf, self.cosmo)\
                         *self._eval_tangential_shear(r_proj, z_cl, z_src=z_inf))**2)
mu = np.zeros_like(r_proj)
for i, r in enumerate(r_proj):
    mu[i] = quad(integrand, z_cl, 1000, r)[0] / quad(z_src, z_cl, 1000)[0]
```

m-aguena:
Hi @marina-ricci, I moved all the z dist functions into `z_distribution.py` and modified the tests accordingly.
One thing left to do is to decide whether the code defaults to Chang 2013 or raises an error when no distribution function is passed. I believe we had a group discussion (that included @combet) and we decided to raise an error, but I am not 100% sure.

hsinfan1996:
I added approximations and exact calculations for `z_src_info=='distribution'`.
The code for exact calculations is not very organized, so it would be appreciated if anyone can help me on that end.
For the approximations, I am able to derive the results of `z_src_info=='distribution'` for reduced tangential shear. Following a similar logic and derivation, `mu = 1 / ((1 - beta_s_mean*kappa_inf)**2 - (beta_s_mean*gamma_inf)**2)` is actually not a bad approximation. There are four lines of multiplicative correction factor to it and the difference compared to the discrete case goes down to ~1%. We can correct `mu` by
```
# exact
mu *= 1-(beta_s_mean*kappa_inf-beta_s_mean*gamma_inf)
mu *= 1-(beta_s_mean*kappa_inf+beta_s_mean*gamma_inf)

# Taylor expansion with up to second-order terms
mu *= 1+(beta_s_mean*kappa_inf-beta_s_mean*gamma_inf)\
       +beta_s_square_mean*kappa_inf**2+beta_s_square_mean*gamma_inf**2\
       -2*beta_s_square_mean*kappa_inf*gamma_inf
mu *= 1+(beta_s_mean*kappa_inf+beta_s_mean*gamma_inf)\
       +beta_s_square_mean*kappa_inf**2+beta_s_square_mean*gamma_inf**2\
       +2*beta_s_square_mean*kappa_inf*gamma_inf
```
However, by playing with the prefactor of the cross terms, we can bring the difference down to <0.1% with
```
# exact
mu *= 1-(beta_s_mean*kappa_inf-beta_s_mean*gamma_inf)
mu *= 1-(beta_s_mean*kappa_inf+beta_s_mean*gamma_inf)

# Taylor expansion with optimized prefactor of cross terms
mu *= 1+(beta_s_mean*kappa_inf-beta_s_mean*gamma_inf)\
       +beta_s_square_mean*kappa_inf**2+beta_s_square_mean*gamma_inf**2\
       # cross terms switches sign and has a different coefficient
       +4*beta_s_square_mean*kappa_inf*gamma_inf
mu *= 1+(beta_s_mean*kappa_inf+beta_s_mean*gamma_inf)\
       +beta_s_square_mean*kappa_inf**2+beta_s_square_mean*gamma_inf**2\
       # cross terms switches sign and has a different coefficient
       -4*beta_s_square_mean*kappa_inf*gamma_inf
```
Of course, we can always calculate and include more higher-order terms if we want to.

m-aguena:
Hi @hsinfan1996, I tidied the code a bit
It is nice to see that the code has better agreement with a 2 order approximation, really shows we are in the right direction.
My suggestion for this use would be to specify the order of the approximation. So we could have something like:

- `approx='order1'` (and also working with `approx='weak lensing'`) : we just compute the first order term
- `approx='order2'`: we also add the 2 order term

We can maybe have a better name for this `order#` argument, but the idea would be to have this mechanism implemented.
@marina-ricci @combet what do you guys think?

m-aguena:
[from 25/11/2022 tag-up]:
- make `approx` just have options like `order1, order2` and add in the documentation which corresponds to know forms (Applegate19, Shrabback18, etc)
- make sure there is a notebook describing the uses of zsrc_info

hsinfan1996:
It turns out the second order term in kappa_inf does not make the approximation better than the 'schrabback18'.
image
For the naming of the approximations, I am not sure if I should call them 'order1', 'order2'  or 'order0', 'order1'. The multiplicative factor in 'schrabback18' is of order 1 (in kappa_inf), so the order of the approximation depends in the multiplicand, which is the 'applegate14'. I call them  'order1' and 'order2' for now so that they are consistent with `eval_magnification` and `eval_magnification_bias`.

hsinfan1996:
> @hsinfan1996 is this PR ready to be reviewed? Can you please change it from draft if it is? Thanks

I think @marina-ricci wants to take a look at the notebook first, but I am happy to mark it as ready since the main components are done.

marina-ricci:
@combet : I just updated the notebook `Example_different_redshift_input_type` that should be seen as finalized.

combet:
Thank you @marina-ricci. 

I'm started reviewing the PR and  I noticed is that the `demo_theory` and `Example3` are not running due to the new names for the redshift distribution function.

clmm/theory/func_layer.py
```diff
@@ -388,76 +405,102 @@ def compute_tangential_shear(r_proj, mdelta, cdelta, z_cluster, z_source, cosmo,
...
+    z_src_info : str, optional
+        Type of redshift information provided, it describes z_source.
+        The following supported options are:
+
+            * 'discrete' (default) : The redshift of sources is provided by `z_source`.
+              It can be individual redshifts for each source galaxy when `z_source` is an array
+              or all sources are at the same redshift when `z_source` is a float.
+
+            * 'distribution' : A redshift distribution function is provided by `z_source`.
+              `z_source` must be a one dimentional function.
```
combet, line 441: dimentional --> dimensional

clmm/theory/func_layer.py
```diff
@@ -388,76 +405,102 @@ def compute_tangential_shear(r_proj, mdelta, cdelta, z_cluster, z_source, cosmo,
...
+    beta_kwargs: None, dict
+        Extra arguments for the `compute_beta_s_mean, compute_beta_s_square_mean` functions.
+        Only used if `z_src_info='distribution'`. Possible keys are:
+
+            * 'zmin' (None, float) : Minimum redshift to be set as the source of the galaxy
+              when performing the sum. (default=None)
+            * 'zmax' (float) : Maximum redshift to be set as the source of the galaxy
+              when performing the sum. (default=10.0)
+            * 'delta_z_cut' (float) : Redshift interval to be summed with $z_cl$ to return
+              $zmin$. This feature is not used if $z_min$ is provided. (default=0.1)
```
combet, line 466: -->  'delta_z_cut' (float) : Redshift cut so that zmin = z_cl - delta_z_cut. delta_z_cut is ignored if z_min is already provided.

clmm/theory/func_layer.py
```diff
@@ -675,78 +761,107 @@ def compute_magnification(r_proj, mdelta, cdelta, z_cluster, z_source, cosmo, de
...
+                \langle \beta_s^2 \rangle = \left\langle \left(\frac{D_{LS}}{D_S}\frac{D_\infty}
+                {D_{L,\infty}}\right)^2 \right\rangle
+
+    approx : str, optional
+        Type of computation to be made for reduced shears, options are:
+
+            * None (default): Full computation is made for each `r_proj, z_source` pair
+              individually. It requires `z_src_info` to be 'discrete' or 'distribution'.
+            * 'weak_lensing' : Uses the weak lensing approximation of the magnification
+              :math:`\my \approx 1 + 2 \kappa`. `z_src_info` must be either 'beta', or
```
combet, line 819: \my \approx --> \mu \approx

clmm/theory/func_layer.py
```diff
@@ -675,78 +761,107 @@ def compute_magnification(r_proj, mdelta, cdelta, z_cluster, z_source, cosmo, de
...
-        If `halo_profile_model=='einasto'`, set the value of the Einasto slope. Option only available
-        for the NumCosmo backend
-    z_src_model : str, optional
-        Source redshift model, with the following supported options:
-            `single_plane` (default) - all sources at one redshift (if
-            `z_source` is a float) or known individual source galaxy redshifts
-            (if `z_source` is an array and `r_proj` is a float);
+        If `halo_profile_model=='einasto'`, set the value of the Einasto slope. Option only
+        available for the NumCosmo backend
+    z_src_info : str, optional
```
combet, line 788: Same docstring comments as before for the whole new redshift-related docstrings.

clmm/theory/func_layer.py
```diff
@@ -788,56 +904,97 @@ def compute_magnification_bias(r_proj, alpha, mdelta, cdelta, z_cluster, z_sourc
...
-    z_src_model : str, optional
-        Source redshift model, with the following supported options:
-            `single_plane` (default) - all sources at one redshift (if
-            `z_source` is a float) or known individual source galaxy redshifts
-            (if `z_source` is an array and `r_proj` is a float);
+            * 'mean' (default)
+            * 'critical'
+            * 'virial'
+
+    z_src_info : str, optional
```
combet, line 928: Same docstring comments as before for the whole new redshift-related docstrings.

clmm/theory/func_layer.py
```diff
@@ -788,56 +904,97 @@ def compute_magnification_bias(r_proj, alpha, mdelta, cdelta, z_cluster, z_sourc
...
+                {D_{L,\infty}}\right)^2 \right\rangle
+
+    approx : str, optional
+        Type of computation to be made for reduced shears, options are:
+
+            * None (default): Full computation is made for each `r_proj, z_source` pair
+              individually. It requires `z_src_info` to be 'discrete' or 'distribution'.
+
+            * 'weak lensing' : Uses the weak lensing approximation of the magnification bias
+              :math:`\mu \approx 1 + 2 \kappa \left(\alpha - 1 \right)`. `z_src_info` must be
```
combet, line 958: \mu --> \delta_\mu (magnification bias here, not magnification)

clmm/theory/parent_class.py
```diff
@@ -587,7 +591,132 @@ def eval_surface_density_2h(self, r_proj, z_cl, halobias=1., lsteps=500):
...
+        r"""Get mean value of the geometric lensing efficicency ratio from typical class function.
+
+        Parameters
+        ----------
+        z_cl : float
+            Galaxy cluster redshift
+        z_src : array_like, float, function
+            Information on the background source galaxy redshift(s). Value required depends on
+            `z_src_info` (see below).
+        z_src_info : str, optional
```
combet, line 604: Same docstring comments as before for the whole new redshift-related docstrings.

clmm/theory/parent_class.py
```diff
@@ -587,7 +591,132 @@ def eval_surface_density_2h(self, r_proj, z_cl, halobias=1., lsteps=500):
...
+        function.
+
+        Parameters
+        ----------
+        z_cl : float
+            Galaxy cluster redshift
+        z_src : array_like, float, function
+            Information on the background source galaxy redshift(s). Value required depends on
+            `z_src_info` (see below).
+        z_src_info : str, optional
```
combet, line 667: Same docstring comments as before for the whole new redshift-related docstrings.

clmm/theory/parent_class.py
```diff
@@ -596,8 +725,48 @@ def eval_tangential_shear(self, r_proj, z_cl, z_src, verbose=False):
             The projected radial positions in :math:`M\!pc`.
         z_cl : float
             Galaxy cluster redshift
-        z_src : array_like, float
-            Background source galaxy redshift(s)
+        z_src : array_like, float, function
+            Information on the background source galaxy redshift(s). Value required depends on
+            `z_src_info` (see below).
+        z_src_info : str, optional
```
combet, line 731: Same docstring comments as before for the whole new redshift-related docstrings.

clmm/theory/parent_class.py
```diff
@@ -642,46 +866,161 @@ def eval_convergence(self, r_proj, z_cl, z_src, verbose=False):
...
+                  tangential shear at each value of `r_proj` is calculated as
+
+                  .. math::
+                      g_t
+                      =\left<\frac{\beta_s\gamma_{\infty}}{1-\beta_s\kappa_{\infty}}\right>
+                      =\frac{\int_{z_{min}}^{z_{max}}\frac{\beta_s(z)\gamma_{\infty}}
+                      {1-\beta_s(z)\kappa_{\infty}}N(z)\text{d}z}
+                      {\int_{z_{min}}^{z_{max}} N(z)\text{d}z}
+
+                * 'order1' : Same as the approach from Weighing the Giants - III (equation 6 in
```
combet, line 989: Same as the approach from --> Same approach as in...

clmm/theory/parent_class.py
```diff
@@ -642,46 +866,161 @@ def eval_convergence(self, r_proj, z_cl, z_src, verbose=False):
...
+                  Applegate et al. 2014; https://arxiv.org/abs/1208.0605). `z_src_info` must be
+                  either 'beta', or 'distribution' (that will be used to compute
+                  :math:`\langle \beta_s \rangle`)
+
+                  .. math::
+                      g_t\approx\frac{\left<\beta_s\right>\gamma_{\infty}}
+                      {1-\left<\beta_s\right>\kappa_{\infty}}
+
+                * 'order2' : Same as the approach from Cluster Mass Calibration at High
```
combet, line 998: Same as the approach from --> Same approach as in...

clmm/theory/parent_class.py
```diff
@@ -735,31 +1086,133 @@ def eval_magnification(self, r_proj, z_cl, z_src, verbose=False):
             The projected radial positions in :math:`M\!pc`.
         z_cl : float
             Galaxy cluster redshift
-        z_src : array_like, float
-            Background source galaxy redshift(s)
+        z_src : array_like, float, function
+            Information on the background source galaxy redshift(s). Value required depends on
+            `z_src_info` (see below).
+        z_src_info : str, optional
+            Type of redshift information provided, it describes z_src.
```
combet, line 1093: Same docstring comments as before for the whole new redshift-related docstrings.

clmm/theory/parent_class.py
```diff
@@ -771,28 +1224,143 @@ def eval_magnification_bias(self, r_proj, z_cl, z_src, alpha):
...
         z_cl : float
             Galaxy cluster redshift
-        z_src : array_like, float
-            Background source galaxy redshift(s)
+        z_src : array_like, float, function
+            Information on the background source galaxy redshift(s). Value required depends on
+            `z_src_info` (see below).
         alpha : float
             Slope of the cummulative number count of background sources at a given magnitude
+        z_src_info : str, optional
```
combet, line 1232: Same docstring comments as before for the whole new redshift-related docstrings.

combet:
> I've started reviewing the PR and I noticed is that the two `demo_theory_functionality*` notebooks and the `Example3` notebook are not running due to the new names for the redshift distribution function.

Now fixed. Thanks @hsinfan1996

tests/test_theory_parent.py
```diff
@@ -104,43 +108,6 @@ def test_instantiate(modeling_data):
...
-
-    reduced_shear = mod.eval_reduced_tangential_shear(r_proj, z_cl, np.repeat(z_src, len(r_proj)), 'applegate14', beta_s_mean, beta_s_square_mean)
-    assert_allclose(reduced_shear, beta_s_test * shear_inf/(1.0 - beta_s_square_test / beta_s_test * convergence_inf), rtol=1.0e-12)
-
-    reduced_shear = mod.eval_reduced_tangential_shear(r_proj, z_cl, z_src, 'schrabback18', beta_s_mean, beta_s_square_mean)
-    assert_allclose(reduced_shear, (1. + (beta_s_square_test / (beta_s_test * beta_s_test) - 1.) * beta_s_test * convergence_inf) * (beta_s_test * shear_inf / (1. - beta_s_test * convergence_inf)), rtol=1.0e-12)
-
-    reduced_shear = mod.eval_reduced_tangential_shear(r_proj, z_cl, np.repeat(z_src, len(r_proj)), 'schrabback18', beta_s_mean, beta_s_square_mean)
-    assert_allclose(reduced_shear, (1. + (beta_s_square_test / (beta_s_test * beta_s_test) - 1.) * beta_s_test * convergence_inf) * (beta_s_test * shear_inf / (1. - beta_s_test * convergence_inf)), rtol=1.0e-12)
-
```
combet, line 143: I don't see these tests moved somewhere else. We don't need them anymore ?

tests/test_theory_parent.py
```diff
@@ -104,43 +108,6 @@ def test_instantiate(modeling_data):
...
-    reduced_shear = mod.eval_reduced_tangential_shear(r_proj, z_cl, z_src, 'applegate14', beta_s_mean, beta_s_square_mean)
-    assert_allclose(reduced_shear, beta_s_test* shear_inf/(1.0 - beta_s_square_test / beta_s_test * convergence_inf), rtol=1.0e-12)
-
-    reduced_shear = mod.eval_reduced_tangential_shear(r_proj, z_cl, np.repeat(z_src, len(r_proj)), 'applegate14', beta_s_mean, beta_s_square_mean)
-    assert_allclose(reduced_shear, beta_s_test * shear_inf/(1.0 - beta_s_square_test / beta_s_test * convergence_inf), rtol=1.0e-12)
-
-    reduced_shear = mod.eval_reduced_tangential_shear(r_proj, z_cl, z_src, 'schrabback18', beta_s_mean, beta_s_square_mean)
-    assert_allclose(reduced_shear, (1. + (beta_s_square_test / (beta_s_test * beta_s_test) - 1.) * beta_s_test * convergence_inf) * (beta_s_test * shear_inf / (1. - beta_s_test * convergence_inf)), rtol=1.0e-12)
-
-    reduced_shear = mod.eval_reduced_tangential_shear(r_proj, z_cl, np.repeat(z_src, len(r_proj)), 'schrabback18', beta_s_mean, beta_s_square_mean)
```
combet, line 141: Just to make sure I followed - these tests have been moved to `test_theory`, right?

combet (changes requested):
Thank you @marina-ricci, @hsinfan1996, @m-aguena, for the work you put in this PR. I think all looks good code wise!
- I left a few comments regarding docstrings, to make things a bit clearer. I can take care of those, but flagged them not to forget. Let me know if you want me to do it.
- I left one question/comment regarding some tests that were moved around, I think. Just want to make sure nothing has disappeared
- I fixed a few typos and added some more text in the new NB and the Example3 NB that were modified by this PR; this was to make some steps a bit clearer to the user. Please take a quick look to confirm this is OK.

combet:
@m-aguena, @hsinfan1996 - Today Build and Check is failing at one of the early steps. Maybe simply waiting will do the trick (it did happen in the past if I recall), but if you  have any insight in case it actually needs fixing, that'd be great :)

hsinfan1996:
Looks like python 3.8 related. Try bumping python version to 3.9.

tests/test_theory_parent.py
m-aguena: yes, just look for tests with `order1`/`order2` in it

combet: Thanks!

combet (approved):
Thanks @hsinfan1996 and @m-aguena - all looks good now :) Happy to approve this massive PR!

combet:
By the way, given that this has changed the API, should we bump the version number to 1.5 when merging this PR and clearly state it in the merge message?

PR #519, opened 2022-06-23
Fix tests for Numpy 1.23 and 1.21

hsinfan1996:


hsinfan1996:
Can I also fix a test that gives an error for NumPy 1.21 here? There is no new code involved, I only need to move one if statement into another to prevent `UnboundLocalError: local variable 'reltol' referenced before assignment`.

marina-ricci:
Thank you very much @hsinfan1996 ! So that solves the current problem in main?

marina-ricci (approved):
This looks good to me. Thank you @hsinfan1996 !

Issue #525, opened 2022-06-29
Improve redshift pdf integration + allow to draw one random z from pdf for DeltaSigma weights computation


combet:
- @payerne noted that the weight computation was very slow compared to his own implementation. This is because we are first interpolating the pdf on a predefined grid to perform the integration. This was done to be able to handle the case where each galaxy has its own pdf binning (which is what mock data does). However, it is unlikely (?) to be the case in real life and it is at least not the case of the DC2 photoz data, where each galaxy pdf is given on the same grid.  So we should probably rethink this by either
  - (a) having a new option that bypasses the interpolation
  - (b) or drop support for non-identical pdf binning
  - (c) and/or really start transitioning to `qp`.

In the interest of time and for this to be used in the DC2 cluster WL mass paper, I'd suggest going with (a) for the time being.

-  Another improvement would be to compute the weights using a redshift point estimate randomly drawn from the pdf. This is what was done in the DES paper and should be faster than accurately integration the photoz pdf, without losing much precision as things are then averaged over thousands of galaxies.

PR #526, opened 2022-06-29
Speed up galaxy weights computation if same binning for all the pdz

combet:
Adresses issue #525.
- [x] Prevent interpolation when it's not needed
- [x] Add method to `GalaxyCluster` class to randomly draw a redshift from the pdz of each galaxy (could be useful for the weights computation). Use example added to demo notebook.


hsinfan1996:
I tried drawing random redshifts from a pdz using `_draw_random_points_from_tab_distribution`, and I think it is working.
1. Since the redshift grid is discrete in pzpdf, specifying xmin (xmax) can only pick up the leftmost (rightmost) bin center that is greater (less) than xmin (xmax). There is a blank between 0.3 to 0.325 in the figure below because the bin center that is closest to xmin=0.3 is 0.325 and the situation is similar for `xmax` if we specify it. To get exact xmin and xmax edges, we will need to interpolate the redshift grid for the two endpoints first (I don't think it will be too slow, but is it necessary?).
2. Calling `_draw_random_points_from_tab_distribution` with `nobj=1` once takes 0.1 to 2.7 microseconds depending on the methods of numerical integration from scipy.integrate. Right now it is `simpson`, which is the slowest. `trapezoid` and `cumulative_trapezoid` give the same results (so `trapezoid` is not visible in the figure) when drawing random samples, but `cumulative_trapezoid` is the fastest of all. From the figure below, `cumulative_trapezoid` is not visibly worse to me, so I think it is probably safe to use `cumulative_trapezoid` instead.

random

combet:
Thanks @hsinfan1996 - I tried it out and all looks good. It's also so much faster!

combet:
@hsinfan1996 - thanks, that's great. I tested it and it works just fine. I just wrapped up that function in new method of the GC object and updated a NB to show the new functionality.

hsinfan1996:
@combet Thank you. The notebook is working nicely. I notice that the new GC method `draw_gal_z_from_pdz` does not have docstrings for the argument `overwrite`.

hsinfan1996:
Tests added. I think it is ready to be reviewed. I also corrected a small error I had made on converting from `np.arange` to `np.linspace`.

combet:
Thanks for the test @hsinfan1996 ! I'm putting this ready for review and tagging @m-aguena, @marina-ricci and @payerne for the review.

clmm/utils.py
```diff
@@ -863,3 +871,76 @@ def _srd_z_distrib(redshift, is_cdf=False):
...
+    Returns
+    -------
+    samples : ndarray
+        Random points following the pdf_tab distribution
+    """
+    x_tab = np.array(x_tab)
+    pdf_tab = np.array(pdf_tab)
+    # Compute the tabulated cumulative distribution
+    if xmin:
+        pdf_tab = pdf_tab[x_tab>=xmin]
```
m-aguena, line 932: these cuts here are throwing data away. They are the responsible for the empty gap at 0.3<x<0.325 in [this](https://github.com/LSSTDESC/CLMM/pull/526#issuecomment-1178487483) fig. I think we sould:

- check if `x_tab.min()<xmin`
  - if False, raise a warning with the value of `x_tab.min()`
  - else, we compute the cdf with the full pdf but only sample for cdf(xsample)>cdf(`xmin`)
  
  same thing for `xmax`. You can do it by just computing the cdf of these values:
  ```
  cdf_xmin, cdf_xmax = 0.0, 1.0
  if xmin or xmax:
      cdf_interp = interp1d(x_tab, cdf, kind='linear')
      if xmin:
          cdf_xmin = cdf_interp(xmin)
      if xmax:
          cdf_xmax = cdf_interp(xmax)
  samples = inv_cdf(np.random.random(nobj)*(cdf_xmax-cdf_xmin)+cdf_xmin)
  ```

m-aguena (changes requested):

m-aguena (approved):

PR #530, opened 2022-07-02
Issue/527/fix compute boost

hsinfan1996:


marina-ricci:
Hi @hsinfan1996 ! Could you provide a bit more description or link the corresponding issue please?

chrisdavies234 (approved):

hsinfan1996:
@chrisdavies234 Thank you for taking a look at this PR.

clmm/utils.py
```diff
@@ -22,11 +22,11 @@ def compute_nfw_boost(rvals, rs=1000, b0=0.1) :
 
     Returns
     -------
-    boost_factors : array_like
+    boost_factors : numpy.ndarray
```
m-aguena, line 25: now these arguments are always being converted to np.arrays, so I guess the old description was better, no?

tests/test_utils.py
```diff
@@ -15,62 +15,59 @@
 def test_compute_nfw_boost() :
     """Test the nfw model for boost factor"""
     # Test data
-    rvals = np.arange(1,11)
+    rvals = np.arange(1,11).tolist()
```
m-aguena, line 18: are these `.tolist()` added here necessary?

m-aguena (commented):
Overall, it looks ok, just some minor comments. I will just run the notebooks to check if everything is ok.

clmm/utils.py
hsinfan1996: I thought here was for the output of the function. Output of the function is always np array regardless of the input type.

tests/test_utils.py
hsinfan1996: It is not necessary. I added it for consistency because all other tests use lists as the input, so we can check if lists are supported. The old tests failed to capture the fact that the relevant functions could not handle lists as the input.

m-aguena:
all notebooks ran ok.

clmm/utils.py
m-aguena: you are right

tests/test_utils.py
m-aguena: ok, good idea

m-aguena (approved):

PR #531, opened 2022-07-06
Issue/529/GC objects weights

hsinfan1996:


combet:
Thanks for looking into this @hsinfan1996. I tested the weights implementation in the GC objects and it works well. 

In your implementation, a list of weights needs to be passed. But the weights should in principle be already stored in a column of `galcat`. So in the OO interface, rather that passing the list of weights I thought that maybe we could simply pass the name of the column where the weights are in the galcat table ? So that instead of calling
`cl0.make_radial_profile('Mpc', cosmo=cosmo, weights=cl0.galcat['w_ls_point'])`
we could simply have
`cl0.make_radial_profile('Mpc', cosmo=cosmo, weights='w_ls_point')` 

What do you think?

hsinfan1996:
> Thanks for looking into this @hsinfan1996. I tested the weights implementation in the GC objects and it works well.
> 
> Just one comment: in your implementation, a list of weights needs to be passed. But the weights should in principle be already stored in a column of `galcat`. So in the OO interface, rather that passing the list of weights I thought that maybe we could simply pass the name of the column where the weights are in the `galcat` table ? So that instead of calling `cl.make_radial_profile('Mpc', cosmo=cosmo, weights=cl.galcat['w_ls'])` we could simply have `cl.make_radial_profile('Mpc', cosmo=cosmo, weights='w_ls')`
> 
> What do you think?

Good. I didn't notice that there is a column for weights. To simplify the process, I added a flag to control if we want to use the default column name given by `compute_galaxy_weights` as the weights. We probably need to check if the column exists when the flag is True.

hsinfan1996:
If we are looking for some `colname` in `galcat` (possibly `profile` and all the other Astropy Table),
using `colname in GalaxyCluster.galcat.columns` (e.g. line 325 of galaxycluster.py) is 6 times faster than
using `colname in GalaxyCluster.galcat.colnames` (e.g.  line 542 of galaxycluster.py).
Would like to know if it is the case for everyone.

combet:
@hsinfan1996 - `coulmns` is also faster than `colnames` at my end and it's more like a factor 10 ! (86 nanosec vs 930 nanosec, timed with %timeit)

combet:
Thanks @hsinfan1996 - I added a demo of the new functionality in the last section of the dataops demo notebook. Could you please check that it looks OK for you and I'll approve the PR.

hsinfan1996:
@combet Thank you. The notebook looks good.

combet (approved):
All looks good!

Issue #524, opened 2022-06-27
Include weights in beta measurements


marina-ricci:
At present, the beta measurements do not include the possibility to account for weight (e.g. shape weights).

eduardojsbarroso:
Hi @marina-ricci . Do you mean that instead of integrating over the redshift interval, we allow the function to receive an array with the shapes `w_i` and use equation (13) from [Cluster Mass Calibration at High Redshift](https://arxiv.org/pdf/1611.03866.pdf)?

marina-ricci:
@eduardojsbarroso : yes, exactly !

m-aguena:
@marina-ricci it is not clear for me where that is going to be applied. Is the idea for the user to compute the values of beta for themselves, and then to use it in the input of the theory functions?

If the answer is positive, than we should add a function to measure beta, as the one we do have makes a theoretical prediction of beta, given a N(z). And I think this function should be located in the `GalaxyCluster` object.

PR #532, opened 2022-07-21
Issue/524/weights in beta

eduardojsbarroso:
Include weights in beta measurements #524


eduardojsbarroso:
@marina-ricci Is this what you were thinking about? Let me know if you want something different.

m-aguena:
[from 25/11/2022 tag-up]: wait for PR #517

eduardojsbarroso:
@m-aguena I think everything is working now. I added the option of the user to provide shape weights for the source galaxies. If you can take a look if this is what you want please.

clmm/utils.py
```diff
@@ -782,8 +782,7 @@ def integrand(z_i, z_cl=z_cl, cosmo=cosmo):
     B_mean = quad(integrand, zmin, zmax)[0] / quad(z_distrib_func, zmin, zmax)[0]
     return B_mean
 
-def compute_beta_s_mean(z_cl, z_inf, cosmo, zmax=10.0, delta_z_cut=0.1, zmin=None,
-                        z_distrib_func=None):
+def compute_beta_s_mean(z_cl, z_inf, cosmo, zmax=10.0, delta_z_cut=0.1, zmin=None, z_distrib_func=None, weights_option=False, z_src = None, shape_weights = None):
```
m-aguena, line 785: I think you can remove the `weights_option` argument and just apply weight when `shape_weights` is provided

m-aguena (changes requested):
Overall is ok. I made comments in specific lines, but they apply to several modifications.

clmm/utils.py
eduardojsbarroso: Done

clmm/utils.py
```diff
@@ -799,36 +798,49 @@ def compute_beta_s_mean(z_cl, z_inf, cosmo, zmax=10.0, delta_z_cut=0.1, zmin=Non
...
-        z_distrib_func = zdist.chang2013
-
-    def integrand(z_i, z_cl=z_cl, z_inf=z_inf, cosmo=cosmo):
-        return compute_beta_s(z_i, z_cl, z_inf, cosmo) * z_distrib_func(z_i)
-
-    if zmin==None:
-        zmin = z_cl + delta_z_cut
-    Bs_mean = quad(integrand, zmin, zmax)[0] / quad(z_distrib_func, zmin, zmax)[0]
+    if weights_option == False:
+        if z_distrib_func == None:
```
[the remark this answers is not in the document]
eduardojsbarroso: Done

clmm/utils.py
```diff
@@ -668,20 +668,22 @@ def compute_beta(z_src, z_cl, cosmo):
...
     -------
-    float
-        Geometric lensing efficicency
+    numpy array
+        Geometric lensing efficicency for each source
     """
-    beta = np.heaviside(z_src-z_cl, 0) * cosmo.eval_da_z1z2(z_cl, z_src) / cosmo.eval_da(z_src)
-    return beta
+    if type(z_src) is not list:
+        z_src = [z_src]
```
eduardojsbarroso, line 684: I don't know if I should keep this test or expect the user to pass a list with just one source. @m-aguena

m-aguena:
@combet @marina-ricci @eduardojsbarroso I was checking this use, and to sum up we have two possible ways of computing `<beta_s>, <beta_s^2>`:

- From N(z): 
image
- From weights:
image

My suggestion here is to just have different functions (ex: `beta_from_distribution`, `beta_from_weights`) as the inputs for each one are completely different.

Then we have two options for how this will be used in the theory functions (ex: `eval_tangential_shear`):

1. The user has to pre-compute `<beta_s>, <beta_s^2>`, and pass it to theory using `z_src=(<beta_s>, <beta_s^2>), z_src_model='beta'`
2. We add the possibility for this computation to be done inside the modeling functions. In this case we have to add an option like `z_src_info='beta_from_weights'` with `z_src=(src_redshifts, src_weights)`

Opinions?

eduardojsbarroso:
> @combet @marina-ricci @eduardojsbarroso I was checking this use, and to sum up we have two possible ways of computing `<beta_s>, <beta_s^2>`:
> 
> * From N(z):
>   image
> * From weights:
>   image
> 
> My suggestion here is to just have different functions (ex: `beta_from_distribution`, `beta_from_weights`) as the inputs for each one are completely different.
> 
> Then we have two options for how this will be used in the theory functions (ex: `eval_tangential_shear`):
> 
> 1. The user has to pre-compute `<beta_s>, <beta_s^2>`, and pass it to theory using `z_src=(<beta_s>, <beta_s^2>), z_src_model='beta'`
> 2. We add the possibility for this computation to be done inside the modeling functions. In this case we have to add an option like `z_src_info='beta_from_weights'` with `z_src=(src_redshifts, src_weights)`
> 
> Opinions?

I think the second option is better than to having the users to compute themselves.

marina-ricci:
> > @combet @marina-ricci @eduardojsbarroso I was checking this use, and to sum up we have two possible ways of computing `<beta_s>, <beta_s^2>`:
> > 
> > * From N(z):
> >   image
> > * From weights:
> >   image
> > 
> > My suggestion here is to just have different functions (ex: `beta_from_distribution`, `beta_from_weights`) as the inputs for each one are completely different.
> > Then we have two options for how this will be used in the theory functions (ex: `eval_tangential_shear`):
> > 
> > 1. The user has to pre-compute `<beta_s>, <beta_s^2>`, and pass it to theory using `z_src=(<beta_s>, <beta_s^2>), z_src_model='beta'`
> > 2. We add the possibility for this computation to be done inside the modeling functions. In this case we have to add an option like `z_src_info='beta_from_weights'` with `z_src=(src_redshifts, src_weights)`
> > 
> > Opinions?
> 
> I think the second option is better than to having the users to compute themselves.

I actually think we can live with option 1 (for which we do not have to make modifications) for the time being. I think it would be cleaner that way.

eduardojsbarroso:
@m-aguena Do you know why is the test failling? The tets work in my computer.

m-aguena:
@eduardojsbarroso yes, it has to do with CCL's new version. PR #593 should fix it

eduardojsbarroso:
@m-aguena  There are a lot of function in `parent_class` that don't have the option `distribution` anymore, while in the notebook they were computed previously. This option should be added to all the functions or should I remove from the notebook?

m-aguena:
> @m-aguena There are a lot of functions in `parent_class` that don't have the option `distribution` anymore, while in the notebook they were computed previously. This option should be added to all the functions or should I remove from the notebook?

@eduardojsbarroso They should be removed from the notebooks. Now, beta should be computed by the user and passed to these function, and this should be reflected on the notebooks

m-aguena (approved):

m-aguena:
@combet @marina-ricci can one of you guys take a final look before I merge this? I would like a second pair of eyes on it since there are some API changes. Thanks!

Issue #501, opened 2022-04-01
CLMM on DES Y1 Data


zhouconghao:
Run CLMM on DES Y1 metacal and redMaPPer catalogs. The product would be a Jupyter notebook.

zhouconghao:
Hi @marina-ricci ! Sorry I missed your message. I didn't make much progress but Shenming and I will try to complete this during the DESC meeting.

PR #535, opened 2022-09-06
Running CLMM on DES data

shenmingfu:
Solving issue #501.
Summarizing SprintDayChicago2022 project https://github.com/LSSTDESC/SprintDayChicago2022/issues/9.


examples/Example5_Fit_Halo_mass_to_DES_data.ipynb
```diff
@@ -0,0 +1,821 @@
...
+      "CPU times: user 4.23 ms, sys: 0 ns, total: 4.23 ms\n",
+      "Wall time: 3.11 ms\n"
+     ]
+    }
+   ],
+   "source": [
+    "%%time\n",
+    "# Assume the downloaded catalog is at this path:\n",
+    "filename = \"../../../ACOS520_DES.csv\"\n",
+    "catalog = filename.replace('.csv', '.pkl')\n",
```
eduardojsbarroso, line 145: @shenmingfu  I'm running the notebook and the query give a `.txt` file, not a `.csv`.

shenmingfu: Yes, the suffix needs to be adjusted manually from .txt to .csv. The text inside is in the csv format.

combet (commented):
Thanks @shenmingfu - it's great to see CLMM applied to DES data!

My main comment concerns cell 12 "Theoretical predictions". With PR #517, CLMM will very soon have functions to compute the beta values and the reduced tangential shear, given those beta values (also with an updated API on how to pass redshift information to the modeling functions).  

PR #517 is almost done, so I suggest holding off just a little while for your PR, so that when #517 is completed, your notebook may be updated to use the new functions in cell 12. I think it is easier this way. Sorry for the short delay :)

shenmingfu:
@combet Thank you. I need some help for using the new functionality in #517 because I'm not familiar with that. Or is it possible to merge this PR first (since only one file changed -- a new notebook) and then update it in #517?

combet:
Acutally, I just realised that the HSC notebook will also need to be updated with the new functions. I suggest we take care of both in this PR (both "theoretical predictions" cells in the DES and HSC notebooks will be the same).

combet:
@shenmingfu - I made a few minor "cosmetic" modifications to the notebook, added a couple of sentences for clarification regarding the shear response, and removed a couple of unnecessary/redundant imports, variables. It was easier to do it directly than ask you to do those minor changes...

Could you please take a look to confirm sure all is in order?

Also, I just approved #517, so last thing to do is update the theory prediction cells in the HSC and DES notebooks (@hsinfan1996 ), and I think that will be it.

shenmingfu:
Thank you @combet. The modifications look good.
Thank you @hsinfan1996. Could you help update the "theoretical predictions" section?

hsinfan1996:
> Thank you @combet. The modifications look good. Thank you @hsinfan1996. Could you help update the "theoretical predictions" section?

No problem.

hsinfan1996:
@shenmingfu I updated some functions. First, I compute `beta_s` using the function in utils.py. Second, the reduced tangential shear profile is calculated with the updated `compute_reduced_tangential_shear` by passing `bs_mean` and `bs2_mean` as a tuple to the function. At this point, everything is the same as what you did. The difference is that our 'order1' approximation (formerly `'applegate14'`) is slightly different from the approximation you used, which I think is from [](https://ui.adsabs.harvard.edu/abs/1997A%26A...318..687S/abstract). I hope the results are still valid after those changes. One last note, I only have HSC data, so I run all the cells in the HSC notebook with no problem. I then copy and paste the changes to the DES notebook without running the notebook, so it would be great if you can run the DES notebook and check if all cells and the results are ok. Thank you.

combet:
Thanks @hsinfan1996 - I successfully ran the DES example and, from the look of it, the results haven't changed (at least not sufficiently for me to notice it) compared to @shenmingfu's original implementation.

combet (approved):
Thank you @shenmingfu and @hsinfan1996!

shenmingfu:
Thank you very much @hsinfan1996 @combet ! Sorry for my late reply. I was busy in the last few days. 
The new notebooks Example 4 and 5 work for me now. Some small things that I noticed: 
1. I needed to update my scipy package to let the Example notebooks run.
2. The notebooks were moved to a new folder, so the paths in the notebooks need to be adjusted a bit.

m-aguena (approved):

PR #536, opened 2022-09-06
Issue #509 zsrc type v2

m-aguena:
PR to show other possibility of #517

### DIFF
--- a/clmm/theory/parent_class.py
+++ b/clmm/theory/parent_class.py
@@ -646,17 +646,9 @@ class CLMModeling:
 
         return self._eval_convergence(r_proj=r_proj, z_cl=z_cl, z_src=z_src)
 
-<<<<<<< HEAD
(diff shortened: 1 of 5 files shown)


clmm/theory/parent_class.py
```diff
@@ -692,36 +712,55 @@ def eval_reduced_tangential_shear(self, r_proj, z_cl, z_src, z_src_model='single
         if self.validate_input:
             validate_argument(locals(), 'r_proj', 'float_array', argmin=0)
             validate_argument(locals(), 'z_cl', float, argmin=0)
-            validate_argument(locals(), 'z_src', 'float_array', argmin=0)
+            if z_src_info_model=='discrete':
```
hsinfan1996, line 715: Would it be `z_src_model` instead of `z_src_info_model`?

PR #545, opened 2022-10-20
Issue/544/test profiles

hsinfan1996:
Add missing tests for Einasto and Herquist profiles. I think these were meant to be done in PR #461 .

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -25,7 +25,7 @@ def modeling_data(request):
     except ValueError:
         pytest.skip(f"Unsupported backend `{param}'.")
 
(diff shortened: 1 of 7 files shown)


hsinfan1996:
Is `sigmac_corr` in line 487 some leftover code? It is exactly 1.

hsinfan1996:
@marina-ricci Do you want to move the Einasto-specific tests in line 310 to 332 to `test_shear_convergence_unittests`? I saw you try to move that part out of where it is, but doing so will cause `UnboundLocalError` for old versions of Numpy. See PR #519 .

combet (approved):
Thank you @hsinfan1996 for this important fix!

PR #547, opened 2022-11-03
Split sigmac function in two functions: standard definition and effective one

combet:
This is to clarify the use of the function(s) that are used to compute the critical surface density. The original function has been split in two:
- `compute_critical_surface_density` (+ corresponding OO method) now refers to the standard definition only
- `compute_critical_surface_density_eff` (+ corresponding OO method) corresponds to the effective definition to be used when using the pdz rather than a redshift point estimate.

Documentation has been updated accordingly.


hsinfan1996:
Need to add two sets of tests in test_dataops.py under def `test_compute_galaxy_weights()`
```
    assert_raises(TypeError, da.compute_galaxy_weights, z_lens, cosmo=None, z_source=None,
                  use_pdz=False, pzpdf=pzpdf, pzbins=pzbins,
                  use_shape_noise=True, shape_component1=None, shape_component2=None,
                  use_shape_error=False, shape_component1_err=None, shape_component2_err=None,
                  is_deltasigma=True, validate_input=True)
    assert_raises(TypeError, da.compute_galaxy_weights, z_lens, cosmo=None, z_source=None,
                  use_pdz=True, pzpdf=pzpdf, pzbins=pzbins,
                  use_shape_noise=False,
                  use_shape_error=False, shape_component1_err=None, shape_component2_err=None,
                  is_deltasigma=True, validate_input=True)
```

combet:
Thanks @hsinfan1996 - I've added the two tests you suggested. I think I am OK with the minimal option, but if we believe we need to test all configurations, this can be done as well - although a bit painful :)

clmm/theory/func_layer.py
```diff
@@ -313,17 +313,40 @@ def compute_surface_density_2h(r_proj, z_cl, cosmo, halobias=1, lsteps=500, vali
...
+    """
+
+    gcm.validate_input = validate_input
+    gcm.set_cosmo(cosmo)
+    sigma_c = gcm.eval_critical_surface_density(z_cluster, z_src=z_source)
+
+    gcm.validate_input = True
+    return sigma_c
+
+def compute_critical_surface_density_eff(cosmo, z_cluster, pzbins=None, pzpdf=None, validate_input=True):
```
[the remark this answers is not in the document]
m-aguena: I agree, those shouldn't have default values

m-aguena:
This should probably wait for PR #517

clmm/theory/func_layer.py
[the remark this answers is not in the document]
combet: Fixed

combet:
@m-aguena, @marina-ricci - Now that #517 is done, this is ready to be re-reviewed.

hsinfan1996 (approved):

clmm/theory/parent_class.py
```diff
@@ -238,12 +238,12 @@ def _eval_critical_surface_density(self, z_len, z_src):
...
     def _eval_tangential_shear(self, r_proj, z_cl, z_src):
         delta_sigma = self.eval_excess_surface_density(r_proj, z_cl)
-        sigma_c = self.eval_critical_surface_density(z_cl, z_src)
+        sigma_c = self.cosmo.eval_sigma_crit(z_cl, z_src)
         return delta_sigma/sigma_c
 
     def _eval_convergence(self, r_proj, z_cl, z_src, verbose=False):
         sigma = self.eval_surface_density(r_proj, z_cl, verbose=verbose)
-        sigma_c = self.eval_critical_surface_density(z_cl, z_src)
+        sigma_c = self.cosmo.eval_sigma_crit(z_cl, z_src)
```
hsinfan1996, line 246: Is there a specific reason not to call `eval_critical_surface_density`?

combet: Hmmm, no not really and I don't recall why I changed it. Maybe because it seems a bit more "native" to call it from the cosmo object than from the modeling object. Actually, we have a lot of redundancy regarding that function, given that `eval_critical_surface_density` and `_eval_critical_surface_density` only call `cosmo.eval_sigma_crit` without adding anything else to it... Maybe we should only keep the one in `Cosmology`?

hsinfan1996: I have no strong opinion on this part, having asked out of curiosity. But I would like to hear what @m-aguena thinks.

combet: OK, so after discussing with @m-aguena yesterday, we agreed that these are probably redundant and that we should only keep the one from the Cosmology class. I will open a new issue to deal with it though. So this PR can be merged as is I think.

m-aguena (approved):

PR #549, opened 2022-11-03
Isse/539/zinf

combet:
This makes `z_inf` an attribute of the CLMM modeling object, to avoid having it hard-coded in the code. The default value is 1000.

### DIFF
--- a/clmm/__init__.py
+++ b/clmm/__init__.py
@@ -15,4 +15,4 @@ from .theory import (
 )
 from . import support
 
-__version__ = '1.4.2'
+__version__ = '1.4.3'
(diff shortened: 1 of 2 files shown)


hsinfan1996 (approved):

m-aguena:
[from 25/11/2022 tag-up]: Accept it as it is and think about implementing it to the functional form later.

hsinfan1996:
For now if users really want to change `z_inf` in the functional interface, they can do `clmm.theory.func_layer.gcm.z_inf = #`

hsinfan1996:
For now if users really want to change `z_inf` in the functional interface, they can do `clmm.theory.func_layer.gcm.z_inf = #`

PR #550, opened 2022-11-04
Issue/462/mock cluster ensemble

eduardojsbarroso:


examples/draw_M_z_from_hmf.ipynb
```diff
@@ -0,0 +1,863 @@
...
+    "\n",
+    "    # Keeps drawing until N points are accepted  \n",
+    "    ran_logm = [] # output list of random numbers  \n",
+    "    ran_z = [] # output list of random numbers  \n",
+    "    while naccept < N:  \n",
+    "        # draw (logm,z) from uniform distribution\n",
+    "        # draw p from uniform distribution\n",
+    "        logm = np.random.uniform(logm_min,logm_max) # x'  \n",
+    "        z = np.random.uniform(zmin,zmax) # x'  \n",
+    "        p = np.random.uniform(pmin,pmax) # y'  \n",
```
eduardojsbarroso, line 348: I changed the pdf to start from pmin instead of 0

examples/draw_M_z_from_hmf.ipynb
```diff
@@ -0,0 +1,863 @@
...
+   "source": [
+    "clusterensemble.make_stacked_radial_profile(tan_component='DS_t', cross_component='DS_x')"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "771c1382",
+   "metadata": {},
+   "source": [
+    "## Covariance (Bootstrap, sample, Jackknife) of the stack between radial bins\n",
```
eduardojsbarroso, line 706: I don't know if we are keeping the covariance analysis on this notebook.

examples/draw_M_z_from_hmf.ipynb
combet: I reverted to 0 as the sanity check was not passing when using pmin. I remember having to convince myself that this was what needed to be used, but cannot think of the reason now. Definitely needs to go back to it, but the behaviour starting from 0 is the right one.

examples/draw_M_z_from_hmf.ipynb
combet: I've left it for the time being, but maybe we can just use one covariance and not compute the three. Let's see with whoever reviews the PR.

examples/draw_M_z_from_hmf.ipynb
m-aguena: If you sort a random number starting from `pmin` points around `P(m, z)=pmin` will never be selected. You want them selected `pmin/pmax` of the time.

m-aguena (approved):

PR #551, opened 2022-11-24
Update GitHub Actions

hsinfan1996:
1. Use `ubuntu-20.04` instead of `ubuntu-latest`, which now refers to Ubuntu 22.04 (was 20.04)
2. Use `actions/checkout@v3` because Node.js 12 (from v2) actions are deprecated according to the warning from GitHub Actions
3. Use `pip install .` instead of `python setup.py install`, which is deprecated
4. `-y` is issued automatically throughout, so we can drop it in the script

### DIFF
--- a/.github/workflows/build_check.yml
+++ b/.github/workflows/build_check.yml
@@ -8,15 +8,14 @@ on:
 
 jobs:
   build-gcc-ubuntu:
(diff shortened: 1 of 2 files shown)


m-aguena (approved):

hsinfan1996:
@m-aguena Now everything works on Ubuntu 22.04 (`ubuntu-latest`)

PR #553, opened 2022-12-19
Update dependencies

hsinfan1996:
Force NumCosmo=0.15.4 (0.16.0 is the latest)

### DIFF
--- a/.github/workflows/build_check.yml
+++ b/.github/workflows/build_check.yml
@@ -39,6 +39,7 @@ jobs:
         pip install .
     - name: Run the unit tests
       run: |
+        conda install numpy=1.23.5
         pip install pytest pytest-cov
         pytest tests/ --ignore=cluster_toolkit/tests --cov=clmm/
       env:
--- a/.github/workflows/build_check.yml
+++ b/.github/workflows/build_check.yml
@@ -39,7 +39,7 @@ jobs:
         pip install .
     - name: Run the unit tests
       run: |
-        conda install numpy=1.23.5
+        conda install -c conda-forge numpy=1.23.5
         pip install pytest pytest-cov
         pytest tests/ --ignore=cluster_toolkit/tests --cov=clmm/
       env:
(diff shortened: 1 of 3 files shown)


hsinfan1996:
We can also relax `'ps_reltol'` in tests/conftest.py to 6e-4 to fix the problem, because all the others tests are not affected by the new version.

hsinfan1996:
Numpy 1.24.0 is now on conda-forge, and it will break Astropy. The Astropy team has pushed a fix, and hopefully there will be a new minor release soon.

m-aguena:
Let's use this patch for now and put a pin for the new astropy version

hsinfan1996:
Sorry, I made it confusing. NumCosmo and Astropy (due to Numpy 1.24.0) are separate issues. Before there is a new minor version for Astropy, we will need to pin NumCosmo=0.15.4 and Numpy=1.23.5.

m-aguena:
@hsinfan1996 how is the build is working without fixing numpy?

hsinfan1996:
> @hsinfan1996 how is the build is working without fixing numpy?

That was the run before Numpy was pushed to conda-forge.

m-aguena:
I see, and a new astropy release will fix this? If it is the case, I say we proceed the way I suggested

hsinfan1996:
> I see, and a new astropy release will fix this? If it is the case, I say we proceed the way I suggested

Yes. Do you mean pinning NumCosmo after Astropy releases a new minor version?

m-aguena:
> > I see, and a new astropy release will fix this? If it is the case, I say we proceed the way I suggested
> 
> Yes. Do you mean pinning NumCosmo after Astropy releases a new minor version?

No, I mean we will be able to unpin the numcosmo and numpy versions after the new astropy mirror, right? We just have to fix them for now

hsinfan1996:
To unpin NumCosmo, we will need to relax the tolerance on power spectrum calculation.

m-aguena:
@hsinfan1996 do you know exactly why the latest version of numcosmo does not pass? Is is an issue on numcosmo or on numpy?

m-aguena (approved):

PR #557, opened 2022-12-21
CCL free einasto alpha

hsinfan1996:


hsinfan1996:
Tests are compatible with CCL<2.6, but the coverage will not be 100% if tested with CCL<2.6.

clmm/theory/ccl.py
```diff
@@ -67,6 +67,9 @@ def __init__(self, massdef='mean', delta_mdef=200, halo_profile_model='nfw',
                           'hernquist': {'truncated': False}}
         self.cor_factor = _patch_rho_crit_to_cd2018(ccl.physical_constants.RHO_CRITICAL)
         self.__mdelta_cor = 0.0 ## mass with corretion for input
+        self._new_version = bool(parse(ccl.__version__) >= parse('2.6'))
+        #if self._new_version:
```
m-aguena, line 71: is this commented part something you forgot?

m-aguena:
> Tests are compatible with CCL<2.6, but the coverage will not be 100% if tested with CCL<2.6.

what is skipped in this case?

hsinfan1996:
> > Tests are compatible with CCL<2.6, but the coverage will not be 100% if tested with CCL<2.6.
> 
> 
> 
> what is skipped in this case?

Code that requires CCL v2.6 to run will not be covered.

clmm/theory/ccl.py
hsinfan1996: I was going to set default alpha_ein to 0.25 (same as the NC backend), but found out it would be difficult for users to debug because the values of the relevant quantities would be different just by installing this new version. I left it there so we can uncomment it once we are ready. I can remove it if it should not be there.

hsinfan1996:
It is necessary to add another check in GitHub Actions to ensure the compatibility of CCL<2.6?

combet (commented):
Thanks @hsinfan1996 - I succesfully ran the new option using CCL v2.6.1!

I noticed that the text before the final cell of the two `demo_theory` notebook (functional and OO) needs to be updated as it still says that setting the value of `alpha_einasto` is only possible for the NumCosmo backend

clmm/theory/func_layer.py
```diff
@@ -61,9 +61,10 @@ def compute_3d_density(r3d, mdelta, cdelta, z_cl, cosmo, delta_mdef=200,
             * 'critical'
             * 'virial'
 
-    alpha_ein : float, optional
-        If `halo_profile_model=='einasto'`, set the value of the Einasto slope. Option only
-        available for the NumCosmo backend
+    alpha_ein : float, 'cosmo'(CCL backend only), optional
+        If `halo_profile_model=='einasto'`, set the value of the Einasto slope.
+        Option only available for the NumCosmo and CCL backends.
+        Set to 'cosmo' to use the cosmology-dependent :math:`\alpha` parameter in the CCL backend.
```
combet, line 67: I find it counter-intuitive that alpha could also be the string `cosmo` for the CCL backend. I wonder whether we should simply mention that not specifying alpha means using the backend default value, which is not the same for CCL or NumCosmo. For CCL this is the cosmology-dependent value and for NumCosmo it is alpha=0.25. 

What do you think?

combet: Actually, trying to implement what I am suggesting, I realise this is actually trickier than it seems as CCL cannot update a parameter to a `None` value (does not give an error, but does not take it into account it seems). If we first call CCL with a user-defined value for alpha, putting back `alpha_ein=None` in the function call does not revert to the CCL default value based on cosmology. This is OK though if calling the function with `alpha_ein=cosmo, as @hsinfan1996 implemented.

I still think it would be good to have a more intuitive behaviour, but we maybe can address this in a new issue?

combet:
> I noticed that the text before the final cell of the two `demo_theory` notebook (functional and OO) needs to be updated as it still says that setting the value of `alpha_einasto` is only possible for the NumCosmo backend

I took care of it :)

m-aguena:
Add comment on the merge message and on documentation about CCL version

combet (changes requested):
Thanks @hsinfan1996 - unfortunately I am still experiencing the behaviour in the functional interface where reverting `alpha_ein` to None, after having specified some value is not taken into account.

Looking at the code, I believe the lines 
```  
if alpha_ein is not None:
     gcm.set_einasto_alpha(alpha_ein)
```
wherever they occur in `func_layer.py` need to be replace with 
```
if halo_profile_model=='einasto' or alpha_ein is not None:
       gcm.set_einasto_alpha(alpha_ein)
```
I did get the right behavior with this looking only at `compute_3d_density`. Would be good if you could check and make the change throughout `func_layer.py`, if you agree with it.

hsinfan1996:
@combet Thanks for catching this and providing a fix.

combet (approved):
Thanks @hsinfan1996! All good :)

m-aguena (approved):

PR #561, opened 2022-12-23
reorder nbs and add OTHERS in website

m-aguena:
Reorder and rename notebooks and add "Other" section on the website


combet:
I just noticed that all the notebooks are now appearing on the public-facing documentation whereas we only had a few selected ones in the past, simply to highlight some key functionalities. Some NB do not really demonstrate CLMM use but instead, e.g., consistency between backends, which are not relevant for new-users. I don't think they should appear on the online documentation.

Might be good to have this as a discussion item, to identify which NB deserve to be in the online doc.

m-aguena:
@combet yes, that was one of the ideas of this issue. The compiled documentation in this PR can be seen [here](https://me.lsst.eu/maguena/clmm/site_temp/html/)

m-aguena:
Minor fixes:

- [ ] Get a better name for "Model a Galaxy Cluster" notebooks

combet (changes requested):
Thanks @m-aguena - the re-organisation looks good to me! Just a couple of things before approving the PR:
- [x] Only a subset of the NB have figures rendered in the new documentation at the moment. I believe we want all figures/tables to be shown in the online doc.
- [x] Rename "Generate a Mock Cluster" --> "Generate mock cluster data/CLMM GalaxyCluster object"?

m-aguena:
we can also just fix issue #564 here.

combet (approved):
Thank you @m-aguena! The doc looks great now :)

PR #570, opened 2023-02-13
Issue/404/galcat pdf improvement

### LINKED ISSUE #404: equal p(z) bins for all sources in mock_data
The current implementation with `photoz_sigma_unscaled` generates a pdf that is sampled differently for every galaxy. From `_compute_photoz_pdfs`:

```python
for row in galaxy_catalog:
    pdf_range = row['pzsigma']*10.
    zmin, zmax = row['z']-pdf_range, row['z']+pdf_range
    zbins = np.arange(zmin, zmax, 0.03)
```
so the result is a column like:
```python
               pzbins [121]              
-----------------------------------------
 -0.9189157769563053 .. 2.681084223043698
-0.8755947879144116 .. 2.7244052120855917
 -0.9013367845355461 .. 2.698663215464457
 -1.0683426683023802 .. 2.531657331697623
 -1.0260346754041412 .. 2.573965324595862
-0.8371474311387707 .. 2.7628525688612324
 -0.7837649451484183 .. 2.816235054851585
 -1.0040270867734482 .. 2.595972913226555
  -1.1684000482540333 .. 2.43159995174597
-0.7198019908543756 .. 2.8801980091456274
                                      ...
-1.3521789249221339 .. 2.2478210750778693
-0.9578447603266889 .. 2.6421552396733143
 -0.9215149025501451 .. 2.678485097449858
-0.9071426552874368 .. 2.6928573447125665
-1.0662076783207386 .. 2.5337923216792646
-1.2636985872843276 .. 2.3363014127156756
 -1.136257420102231 .. 2.4637425798977723
 -0.7774128187340363 .. 2.822587181265967
-1.0730126785251919 .. 2.5269873214748113
-0.9649808022321504 .. 2.6350191977678525
-1.3844228023509195 .. 2.2155771976490835
```
This is both very inconvenient, because the user then has to interpolate if they want to combine the pdfs, and a big waste of space.

I propose to change this for a single binning scheme that depends on `(zsrc_min,zsrc_max)` and a new optional argument `zsrc_bin_width` or similar, so that the binning can be stored as a separate attribute of `GalaxyCluster`, e.g., `zsrc_pdf_centers` or just `zsrc_bins`. See related suggestion in #401.

m-aguena:
Make pdf have single binning
Closes #404


combet (commented):
Thanks @m-aguena. I did a quick first pass and noticed that the 2`mock_ensemble` notebooks and the `deltasigma_weights` notebook are failing. This is because to compute the "effective sigma_c", the function (which will be split in two in #547) still expects a column named `pzbins` if `use_pdz=True` in the call.

So we probably need an extra condition in there that check the value of the new `pzpdf_info` metadata to access the bins properly.

clmm/support/mock_data.py
```diff
@@ -404,17 +457,32 @@ def _compute_photoz_pdfs(galaxy_catalog, photoz_sigma_unscaled):
...
+            galaxy_catalog['ztrue'][:,None],
+            galaxy_catalog['pzsigma'][:,None])
+    elif galaxy_catalog.pzpdf_info['type']=='individual_bins':
+        galaxy_catalog['pzbins'] = [
+            np.arange(zmin, zmax, pz_bin_width)
+            for zmin, zmax in zip(
+                galaxy_catalog['ztrue']-10.*galaxy_catalog['pzsigma'],
+                galaxy_catalog['ztrue']+10.*galaxy_catalog['pzsigma'])]
+        galaxy_catalog['pzpdf'] = [gaussian(row['pzbins'], row['ztrue'], row['pzsigma'])
+                                   for row in galaxy_catalog]
```
combet, line 478: @m-aguena - Replacing `ztrue` by `z` in this section allows the recover the original behaviour from the `main` branch for the `demo_compute_deltasigma_weights` notebook (see my post on Slack for the figures). I think we need to discuss what implementation make more sense.

m-aguena: @combet I centered the pdzs o `z` back again. I am just wondering if the figure of the weights on the notebook should be made as a function of `z` or `ztrue`

combet: Thanks @m-aguena - Showing the scattered version w.r.t `ztrue` allows us to highlight that two galaxies with the same redshift won't necessary pass a given 'probability cut' of being a background galaxy the same way. I think it has a bit more information than showing it w.r.t to `z` where, by construction of the mock data pdf, 2 galaxies with the same point-estimate redshift will behave the same way in that figure, as we gave them the same pdz. Is that making sense?

m-aguena: @combet yes, I see your point. I decided to update the `demo_compute_deltasigma_weights` with more details.
The PR is ready to be reviewed now. Thanks!

combet (approved):
Thanks @m-aguena! Updating the notebook makes things clearer. All looks good to me :)

PR #579, opened 2023-03-17
Fix CCL 2.6.2-dev7 

hsinfan1996:
Now CCL only allows modifying class attributes by public functions. To fix instantiating a modeling object, we will need to call either `ccl.UnlockInstance` or  `__init__` again.

### DIFF
--- a/clmm/__init__.py
+++ b/clmm/__init__.py
@@ -15,4 +15,4 @@ from .theory import (
 )
 from . import support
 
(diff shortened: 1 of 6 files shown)


hsinfan1996:
We can also use CCL v2.6.1 for GH Actions and worry about this problem after CCL v3 releases.

m-aguena (approved):

PR #587, opened 2023-04-03
Issue/574/auto formatting

m-aguena:
Use of black and pylint


hsinfan1996:
I know this PR is still in progress, but I notice there are `""" "` in 26 places in the opening quotation marks of the docstrings. They are in `clmm/theory/ccl.py`, `clmm/theory/cluster_toolkit.py` and `clmm/theory/numcosmo.py`.

hsinfan1996:
I usually issue `python -m black .` from the CLMM root directory for convenience, but black fails to format `notebook_install.py` because line 15 and 16 start with a `!`. I wonder if we can use `os.system` instead to run the commands.

m-aguena:
> I know this PR is still in progress, but I notice there are `""" "` in 26 places in the opening quotation mark of the docstrings. They are in `clmm/theory/ccl.py`, `clmm/theory/cluster_toolkit.py` and `clmm/theory/numcosmo.py`. Should we fix them?

Nice catch! Yeah, we want to fix those

m-aguena:
> I usually issue `python -m black .` from the CLMM root directory for convenience, but black fails to format `notebook_install.py` because line 15 and 16 start with a `!`. I wonder if we can use `os.system` instead to run the commands.

I am not sure if `os.system` does the job properly. But it would be nice to have a better solution there

hsinfan1996:
> > I usually issue `python -m black .` from the CLMM root directory for convenience, but black fails to format `notebook_install.py` because line 15 and 16 start with a `!`. I wonder if we can use `os.system` instead to run the commands.
> 
> I am not sure if `os.system` does the job properly. But it would be nice to have a better solution there

The '!' symbol is an ipython feature, not python. Python uses `os.system` or `os.popen` to execute shell commands.

Actually, I would just remove `notebook_install.py` because the way we call `install_clmm_pipeline` in the example notebooks will not install clmm for the users (The commands will not be executed because `upgrade=False`). It is just a thought and do correct me if I did not use `os.system` or `notebook_install.py` properly.

m-aguena:
> > > I usually issue `python -m black .` from the CLMM root directory for convenience, but black fails to format `notebook_install.py` because line 15 and 16 start with a `!`. I wonder if we can use `os.system` instead to run the commands.
> > 
> > 
> > I am not sure if `os.system` does the job properly. But it would be nice to have a better solution there
> 
> The '!' symbol is an ipython feature, not python. Python uses `os.system` or `os.popen` to execute shell commands.
> 
> Actually, I would just remove `notebook_install.py` because the way we call `install_clmm_pipeline` in the example notebooks will not install clmm for the users (The commands will not be executed because `upgrade=False`). It is just a thought and do correct me if I did not use `os.system` or `notebook_install.py` properly.

Actually I think this bit is unnecessary. We should just remove this `notebook_install.py` from the code, the user just have to have clmm installed before using the notebooks

clmm/theory/ccl.py
```diff
@@ -50,101 +39,116 @@ class CCLCLMModeling(CLMModeling):
...
         self.set_halo_density_profile(halo_profile_model, massdef, delta_mdef)
         self.set_cosmo(None)
 
-
     # Functions implemented by child class
 
-
     def _update_halo_density_profile(self):
-        """"updates halo density profile with set internal properties"""
+        """ "updates halo density profile with set internal properties"""
```
hsinfan1996, line 89: The opening quotation marks.

CONTRIBUTING.md
```diff
@@ -64,14 +64,14 @@ To review an open pull request submitted by another developer, there are several
...
-3. **Code formatting:** Make sure the code formatting is up to standards, you can do this by running a linter on any new or changed files `pylint {filename}`. This will take a look at the file and identify any style problems. If there are only a couple, feel free to resolve them yourself, otherwise leave a comment in your review that the author should perform this step.
+3. **Code formatting:** Make sure the code formatting is up to standards, you can do this by running a linter on any new or changed files `pylint {filename}`. This will take a look at the file and identify any style problems. If there are only a couple, feel free to resolve them yourself, otherwise leave a comment in your review that the author should perform this step. **Tips for formatting:** CLMM is setup to be formatted automatically with the `black` package (can be installed via `pip install black`). Just run `black {filename}` for specific file or `black .` iside the main CLMM directory for the whole repo. **Bonus tip:** This can also be done for notebooks, just install the notebook extention: `pip install "black[jupyter]"`.
```
combet, line 67: iside --> inside

clmm/cosmology/ccl.py
```diff
@@ -28,25 +28,32 @@ def __init__(self, **kwargs):
...
+        self.cor_factor = _patch_rho_crit_to_cd2018(ccl.physical_constants.RHO_CRITICAL)
 
     def _init_from_cosmo(self, be_cosmo):
-
         assert isinstance(be_cosmo, ccl.Cosmology)
         self.be_cosmo = be_cosmo
 
     def _init_from_params(self, H0, Omega_b0, Omega_dm0, Omega_k0):
-
+        # pylint: disable=arguments-differ
```
combet, line 43: Is this commented line needed?

clmm/theory/ccl.py
```diff
@@ -50,101 +39,116 @@ class CCLCLMModeling(CLMModeling):
...
-        if parse(ccl.__version__) >= parse('2.6.2dev7'):
+        if parse(ccl.__version__) >= parse("2.6.2dev7"):
             ccl.UnlockInstance.Funlock(type(self.mdef), "_concentration_init", True)
         # setting concentration (also updates hdpm)
-        self.cdelta = self.cdelta if self.hdpm else 4.0 # ccl always needs an input concentration
+        self.cdelta = self.cdelta if self.hdpm else 4.0  # ccl always needs an input concentration
 
     def _get_concentration(self):
-        """"get concentration"""
+        """ "get concentration"""
```
combet, line 99: Same to what Hsin mentioned above. Plus several more instances below.

combet (commented):
If I'm not mistaken @m-aguena, you have split `utils.py` into a directory with different files, depending on what type of utils they provided. This is very nice! But now that there is a `utils/redshift.py` module, I'm wondering whether `clmm/z_distribution.py` should be merged into `redshift.py`.

clmm/theory/ccl.py
m-aguena: done

CONTRIBUTING.md
m-aguena: done

clmm/theory/ccl.py
m-aguena: done

clmm/cosmology/ccl.py
m-aguena: yes, they are to tell pylint yo skip a specific check at a specific location. There are some cases where we do want to have code that differs from pylint's formatting, like having arguments with Upper and Lower case (`Omega_dm0`) and with less than three characters (`H0`)

clmm/dataops/__init__.py
```diff
@@ -13,7 +13,10 @@
     validate_argument,
     _validate_ra,
     _validate_dec,
+)
+from ..redshift import (
```
hsinfan1996, line 17: from ..redshift.tools import (

clmm/cosmology/parent_class.py
```diff
@@ -2,7 +2,8 @@
 """
 # CLMM Cosmology object abstract superclass
 import numpy as np
-from ..utils import validate_argument, compute_for_good_redshifts
+from ..utils import validate_argument
+from ..redshift import compute_for_good_redshifts
```
hsinfan1996, line 6: from ..redshift.tools import compute_for_good_redshifts

clmm/theory/parent_class.py
```diff
@@ -20,10 +20,12 @@
 )
 from ..utils import (
     validate_argument,
-    _integ_pzfuncs,
     compute_beta_s_mean,
     compute_beta_s_square_mean,
     compute_beta_s_func,
+)
+from ..redshift import (
```
hsinfan1996, line 27: from ..redshift.tools import (

hsinfan1996:
> @hsinfan1996 I had forgotten to push the `__init__.py` file

Ah ok. That makes sense.

hsinfan1996:
The latest commit to pyccl master removed `_concentration_init`.

combet (approved):
Looks good - thanks and sorry for having forgotten about this review.

Issue #572, opened 2023-02-28
Standardize redshift use in dataops


m-aguena:
Many functions on `dataops` still use  `z_source, use_pdz` as arguments. It should be updated to `z_src, z_src_info`

PR #596, opened 2023-05-25
Issue/572/dataops redshift

m-aguena:
Closes Issue #572 .
In the end, most functions could just change their input to `sigma_c` instead of implementing `z_src, z_src_info`.


clmm/dataops/__init__.py
```diff
@@ -182,58 +152,32 @@ def compute_tangential_and_cross_components(
...
-                    "To compute DeltaSigma using the redshift pdz of the sources, "
-                    "please provide a "
-                    "i) cosmology, ii) lens redshift, iii) source redshift bins and"
-                    "iv) source redshift pdf"
-                )
-
-            sigma_c = compute_critical_surface_density_eff(
-                cosmo, z_lens, pzbins=pzbins, pzpdf=pzpdf
-            )
+    # If the is_deltasigma flag is True, multiply the results by Sigma_crit.
```
hsinfan1996, line 156: Isn't `is_deltasigma` flag removed?

clmm/dataops/__init__.py
```diff
@@ -182,58 +152,32 @@ def compute_tangential_and_cross_components(
...
+    tangential_comp *= sigma_c
+    cross_comp *= sigma_c
```
hsinfan1996, line 158: `tangential_comp *= sigma_c if sigma_c !=1.0 else tangential_comp` is faster.

clmm/dataops/__init__.py
```diff
@@ -242,12 +186,12 @@ def compute_background_probability(
     """
     if validate_input:
         validate_argument(locals(), "z_lens", float, argmin=0, eqmin=True)
-        validate_argument(locals(), "z_source", "float_array", argmin=0, eqmin=True, none_ok=True)
+        validate_argument(locals(), "z_src", "float_array", argmin=0, eqmin=True, none_ok=True)
 
     if use_pdz is False:
```
hsinfan1996, line 191: There is no `use_pdz` flag.

clmm/dataops/__init__.py
```diff
@@ -182,58 +152,32 @@ def compute_tangential_and_cross_components(
...
-    z_lens, z_source=None, use_pdz=False, pzpdf=None, pzbins=None, validate_input=True
+    z_lens, z_src=None, use_pdz=False, pzpdf=None, pzbins=None, validate_input=True
 ):
-    r"""Probability for being a background galaxy
+    r"""Probability for being a background galaxy, defined by:
+
+        .. math::
+            P(z_s > z_l) = \int_{z_l}^{+\infty} dz_s p_{\text{photoz}}(z_s),
+
+    when the photometric probability density functions ($p p_{\text{photoz}}(z_s)$) are provided.
```
hsinfan1996, line 171: Need to use :math:``

clmm/galaxycluster.py
```diff
@@ -276,8 +276,6 @@ def compute_tangential_and_cross_components(
             ra_lens=self.ra,
             dec_lens=self.dec,
             geometry=geometry,
-            is_deltasigma=is_deltasigma,
```
hsinfan1996, line 279: Comment on the docstrings in line 221, where it writes calls `clmm.dataops.compute_tangential_and_cross_components` with the argument `is_deltasigma`. The docstrings  should be updated.

examples/demo_compute_deltasigma_weights.ipynb
hsinfan1996, line 1: Cell 8: `sigma_c` should be the first argument. The other argument would be `use_shape_noise=False`.
Cell 10: `pzbins=noisy_data_z.pzpdf_info['zbins']` instead of `pzbins=noisy_data_z["pzbins"]`
Cell 11: Same as Cell 8.

clmm/dataops/__init__.py
```diff
@@ -105,30 +98,10 @@ def compute_tangential_and_cross_components(
...
-    cosmo: clmm.Cosmology, optional
-        Required if `is_deltasigma` is True and `sigma_c` not provided.
-        Not used if `sigma_c` is provided.
-    z_lens: float, optional
-        Redshift of the lens, required if `is_deltasigma` is True and `sigma_c` not provided.
-        Not used if `sigma_c` is provided.
-    z_source: array, optional
-        Redshift of the source, required if `is_deltasigma` is True and `sigma_c` not provided.
-        Not used if `sigma_c` is provided or `use_pdz=True`.
     sigma_c : float, optional
```
hsinfan1996, line 101: The type of `sigma_c` can also be numpy.ndarray, but probably not array_like.

clmm/galaxycluster.py
hsinfan1996, line 279: Comment on line 221 (docstrings): `is_deltasigma` is removed.

clmm/dataops/__init__.py
```diff
@@ -152,10 +125,7 @@ def compute_tangential_and_cross_components(
         validate_argument(locals(), "shear1", "float_array")
         validate_argument(locals(), "shear2", "float_array")
         validate_argument(locals(), "geometry", str)
-        validate_argument(locals(), "is_deltasigma", bool)
-        validate_argument(locals(), "z_lens", float, argmin=0, eqmin=True, none_ok=True)
-        validate_argument(locals(), "z_source", "float_array", argmin=0, eqmin=True, none_ok=True)
-        validate_argument(locals(), "sigma_c", "float_array", none_ok=True)
+        validate_argument(locals(), "sigma_c", "float_array")
```
hsinfan1996, line 128: Related to the previous comment: `"float_array"` will let a list pass the validation, but it will lead to TypeError at a later stage.

clmm/dataops/__init__.py
```diff
@@ -257,76 +201,50 @@ def compute_background_probability(
...
 
-        If the redshift pdf of each source, :math:`p_{\text{photoz}}(z_s)`, is known,
-        the weights are computed instead as
 
         .. math::
-            w_{ls, \text{geo}} = \left[\int_{\delta + z_l} dz_s p_{\text{photoz}}(z_s)
-            \Sigma_c(\text{cosmo}, z_l, z_s)^{-1}\right]^2
+            w_{ls, \text{geo}} = \Sigma_{\rm crit}^{\rm eff}(z_l, z_{\text{src}})^{-2}
+            = \left[\int_{\delta + z_l} dz_s p_{\text{photoz}}(z_s)
+            \Sigma_{\rm crit}(z_l, z_s)^{-1}\right]^2
```
hsinfan1996, line 231: `\rm`'s

clmm/dataops/__init__.py
m-aguena: yes, this is a leftover

clmm/dataops/__init__.py
```diff
@@ -257,76 +201,50 @@ def compute_background_probability(
...
 
-    1. The geometrical weight :math:`w_{ls, \text{geo}}` depends on lens and source redshift
-    information. When considering only redshift point estimates, the weights read
+    1. When computed for tangential shear, the geometrical weights :math:`w_{ls, \text{geo}}` are
+    equal to 1. When compute for $\Delta \Sigma$, it depends on lens and source redshift information
+    via the critical surface density. It can be expressed as:
 
         .. math::
-            w_{ls, \text{geo}} = \Sigma_c(\text{cosmo}, z_l, z_{\text{src}})^{-2}\;.
+            w_{ls, \text{geo}} = \Sigma_{\rm crit}(z_l, z_{\text{src}})^{-2}\;.
```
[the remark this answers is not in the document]
m-aguena: done

clmm/dataops/__init__.py
m-aguena: I think we can keep this function as it is

hsinfan1996: Yes. Just saw that `use_pdz` was still an argument of the function. It is just not in the docstrings.

examples/demo_compute_deltasigma_weights.ipynb
m-aguena: solved

hsinfan1996 (approved):
https://github.com/LSSTDESC/CLMM/pull/596/files#diff-b0d44342af6305abba113e4054806c63583e6cd75effe601d322c6e56709e6afR227
This line of the docstrings is not correct. Now `is_deltasigma` flag is removed from `clmm.dataops.compute_tangential_and_cross_components`

m-aguena:
@combet @marina-ricci can one of you take a final look before I merge this? I would like a second pair of eyes on it since there are some API changes. The main changes here are on ` dataops/__init__.py` and on `galaxycluster.py
` (and their tests), most changes in the code are just replacing `z_source->z_src`. Thanks!

clmm/dataops/__init__.py
```diff
@@ -105,30 +96,10 @@ def compute_tangential_and_cross_components(
...
-    pzpdf : array, optional
-        Photometric probablility density functions of the source galaxies, required if
-        `is_deltasigma=True` and `use_pdz=True` and `sigma_c` not provided.
-    pzbins : array, optional
-        Redshift axis on which the individual photoz pdf is tabulated, required if
-        `is_deltasigma=True` and `use_pdz=True` and `sigma_c` not provided.
+    sigma_c : None, array_like
+        Critical (effective) surface density in units of :math:`M_\odot\ Mpc^{-2}`.
+        If equals to None, shear components are returned, else :math:`\Delta \Sigma` components
+        are returned.
```
combet, line 102: I like the simplicity brought by having the user previously compute sigma_c, then to pass this to all the functions. But I'm wondering if 'hiding' the decision of computing DeltaSigma instead of shear in the sigma_c keyword may yield some confusion. An option would be to keep a `is_deltasigma` flag for the user to specify, and if `True` check that the user has indeed provided sigma_c. Just a suggestion though.

clmm/dataops/__init__.py
```diff
@@ -257,76 +200,51 @@ def compute_background_probability(
 
 
 def compute_galaxy_weights(
-    z_lens,
-    cosmo,
-    z_source=None,
-    use_pdz=False,
-    pzpdf=None,
-    pzbins=None,
+    sigma_c=None,
```
combet, line 203: Same as above - still request the user to pass sigma_c, but have an explicit flag for them to say they want DeltaSigma weights and not shear weights?

combet:
@m-aguena - thank you for pinging me again on this PR. I took a look this morning and left a couple of comments above. Let me know what you think. Also have a follow up question below (that could be addressed in a subsequent issue/PR).

You implemented this new approach for the functional interface only. The methods `compute_galaxy_weights` and `compute_tangential_and_cross_components` of the `GalaxyCluster` class still require the `use_pdf` and `is_deltasigma` flags. Should they be homogenized as well, i.e the user should have added a `sigma_crit` column beforehand and these methods would just check whether it exists or not ?

m-aguena:
> @m-aguena - thank you for pinging me again on this PR. I took a look this morning and left a couple of comments above. Let me know what you think. Also I have a follow up question below (that could be addressed in a subsequent issue/PR).
> 
> You implemented this new approach for the functional interface only. The methods `compute_galaxy_weights` and `compute_tangential_and_cross_components` of the `GalaxyCluster` class still require the `use_pdf` and `is_deltasigma` flags. Should they be homogenized as well, i.e the user should have added a `sigma_crit` column beforehand and these methods would just check whether it exists or not (probably still keep a `is_deltasigma` flag) ?

@combet, thanks for the comments.

With regards to `is_deltasigma`, I agree it can be more friendly to have them on the functional interface too. The `galaxycluster` object currently computes and adds the `sigma_c` (or `sigma_c_eff`) column automatically if it is not there already when `is_deltasigma=True`. Is your suggestion to return an error when  `sigma_c` (or `sigma_c_eff`) is missing and force the user to add it by hand?

The argument `use_pdf` does not make sense in the functional form as that information is already completely contained in the input `sigma_c`. I agree that it makes the API to be different from the `galaxycluster` object, but I do not see a clear solution. Maybe we could change the argument in `galaxycluster` to something like `sigma_c_colname`, and the options would be `'sigma_c'` or `'sigma_c_eff'`. What do you think?

combet:
> With regards to `is_deltasigma`, I agree it can be more friendly to have them on the functional interface too. The `galaxycluster` object currently computes and adds the `sigma_c` (or `sigma_c_eff`) column automatically if it is not there already when `is_deltasigma=True`. Is your suggestion to return an error when `sigma_c` (or `sigma_c_eff`) is missing and force the user to add it by hand?

Yes, I thought it would be more consistent with what we now have in the functional interface. i.e., the user need to make sure they have pre-computed `sigma_c`. For the OO interface, that would mean they have to explicitly use the `add_critical_surface_density` method, to ensure `galcat` has a sigma_c column. But there might be some side effects to do that, I haven'thought of every possible scenario. 

This could/should be kept for another issue/PR until we carefully think about it. But I think it would be nice to try to keep a good symmetry between the functional and OO interface.

m-aguena:
> > With regards to `is_deltasigma`, I agree it can be more friendly to have them on the functional interface too. The `galaxycluster` object currently computes and adds the `sigma_c` (or `sigma_c_eff`) column automatically if it is not there already when `is_deltasigma=True`. Is your suggestion to return an error when `sigma_c` (or `sigma_c_eff`) is missing and force the user to add it by hand?
> 
> Yes, I thought it would be more consistent with what we now have in the functional interface. i.e., the user need to make sure they have pre-computed `sigma_c`. For the OO interface, that would mean they have to explicitly use the `add_critical_surface_density` method, to ensure `galcat` has a sigma_c column. But there might be some side effects to do that, I haven't thought about it very long :) .
> 
> This could/should be kept for another issue/PR until we carefully think about it. But I think it would be nice to try to keep a good symmetry between the functional and OO interface.

I agree that this is a good idea, and also to put it in another issue so we can think more carefully about it and also to move on with this one.

BTW, I just implemented the changes for the `is_sigmacrit` argument in the functional interface.

clmm/utils/validation.py
```diff
@@ -207,3 +207,20 @@ def _validate_dec(loc, dec_name, is_array):
...
+    ----------
+    is_deltasigma: bool
+        If `False`, values are computed for shear, else they are computed for :math:`\Delta \Sigma`.
+    sigma_c : None, array_like
+        Critical (effective) surface density in units of :math:`M_\odot\ Mpc^{-2}`.
+    """
+    if is_deltasigma and sigma_c is None:
+        raise TypeError("sigma_c (=None) must be provided when is_deltasigma=True")
+    if not is_deltasigma and sigma_c is not None:
+        raise TypeError(f"sigma_c (={sigma_c}) be provided when is_deltasigma=False")
```
combet, line 226: I'd suggest only warning the user that sigma_c is ignored if is_deltasigma=False

```
    if not is_deltasigma and sigma_c is not None:
        warnings.warn(f"sigma_c (={sigma_c}) is ignored when is_deltasigma=False")

```

combet (commented):
Thanks @m-aguena. Playing the `demo_compute_weights` notebook gave errors in cell 9 and 13 because of the new `_validate_is_deltasigma_sigma_c` check. I think the calls in these cells now need to explicitly have a `is_deltasigma=True`.

m-aguena:
> is_deltasigma=True

Thanks @combet , good catch!
I updated the notebook now.

combet (approved):
Thanks @m-aguena. All looks good to me now!

PR #615, opened 2024-01-07
Updated NumCosmo usage due to deprecation of old constructors.

vitenti:


hsinfan1996 (approved):

hsinfan1996:
Example2 would fail silently on my machine at cell 22 under the section `To investigate further the results we make a MCMC analysis below.`

m-aguena (approved):

PR #617, opened 2024-02-20
Update required Scipy version

hsinfan1996:
#615 needs to be merged first in order not to make changes to the CI script.


m-aguena:
@hsinfan1996 is ccl going to require this newer scipy version too?

hsinfan1996:
> @hsinfan1996 is ccl going to require this newer scipy version too?

No, CCL does not depend on scipy. But ccl 3.0.1 is going to release soon and the current mechanism for checking the version number will fail. I think it is quicker to fix it in this PR. I can do it in a separate PR if it is a proper way.

hsinfan1996:
`integrate.simps` has been deprecated and will be removed in Scipy v1.14.

hsinfan1996:
There is one more `simps` here, but I did not change it since it was commented out.
https://github.com/LSSTDESC/CLMM/blob/714b92d9343e19ff0ac9b7555a8c113847e6bbb4/clmm/utils/statistic.py#L213

m-aguena (approved):

PR #620, opened 2024-03-05
Prepare for new CCL release

hsinfan1996:
Fix how max supported CCL version is checked.

### DIFF
--- a/clmm/theory/ccl.py
+++ b/clmm/theory/ccl.py
@@ -13,9 +13,8 @@ from .parent_class import CLMModeling
 # Check which versions of ccl are currently supported
 from . import _ccl_supported_versions
 
-if parse(ccl.__version__) < parse(_ccl_supported_versions.VMIN) or parse(ccl.__version__) > parse(
-    _ccl_supported_versions.VMAX
-):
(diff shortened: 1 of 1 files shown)


hsinfan1996:
Note: origin of the issue `parse("3")` is identical to `parse("3.0.0")`.

hsinfan1996:
pyccl 3.0.1 and 3.0.2 is released.

hsinfan1996:
> Is there a reason you're installing CCL from source? It is available in conda-forge, and much quicker to install that way!

I think it is partly because we want to check the dev versions, and in the CI script, we don't need to specify what version to install.

hsinfan1996:
conda will be broken after installing cmake due to an upstream issue https://github.com/libarchive/libarchive/pull/1976, so move up the installation of sphinx prereqs.

combet:
@hsinfan1996 - I see you've made further changes here that should work for all CCL v3.x.x, right? Thank you! Is this ready for review/to be merged?

hsinfan1996:
@combet Yes, this is ready for review.

combet (approved):
Thank you! All looks good and it installed and ran happily at my end with a conda installed CCL v3.0.2.

m-aguena (approved):

Issue #588, opened 2023-04-05
Add miscentering to the theory module


combet:
Miscentered profiles (single cluster or stacked) should be added to the theory module. See e.g. the [cluster-toolkit documentation](https://cluster-toolkit.readthedocs.io/en/latest/source/miscentering.html).

- for a single cluster --> new R_mis parameter to quantify the miscentering + 2 integrations
- for a stack --> needs extra integration over a miscentering distribution 

Is it better to create new 'miscentered' functions or have extra keywords to the existing ones? What precision are we happy to live with. If sticking with scipy.integrate, we'll have to sacrifice a bit accuracy for speed I believe...

combet:
Just pushed initial notebook that explores the set of integrals that need to be computed.

hsinfan1996:
I made some optimization to the integrals and explored the possibilities of using different profiles and CCL backend. The results are summarized in the notebook.

AnnaNiemiec:
I have added an option to compute Sigma mean miscentered with dblquad instead of two nested quads. Testing this on the `demo_theory_functionality_oo.ipynb` notebook, it seems to increase the speed x2.
However, the two variants (simple quads and dblquad) appear much slower in this notebook, as compared to the `explore_miscentering_theory.ipynb` notebook.
For an NFW profile, I obtain the following times : 

 - `explore_miscentering_theory.ipynb` :
backend independant : 2s
backend independant + dblquad : 0.3s

 - `demo_theory_functionality_oo.ipynb` :
backend independant : 12s
backend independant + dblquad : 5s

Not sure what is causing this difference

PR #622, opened 2024-05-07
Theory prediction for miscentered haloes

combet:
Add functionality to compute profile predictions for miscentered haloes, based on tests performed in the exploratory notebook `explore_miscentering_theory.ipynb`
- new functionality added to the theory module
- the theory demo notebooks (functional interface and OO) updated accordingly
- new tests implemented


combet:
@m-aguena, @marina-ricci  - following on the v2.0 release discussion and the idea of having miscentered profiles predictions part of the release, I'm thinking one compromise would be to implement the miscentered profile for a single halo (i..e not integrated over a miscentering distribution). This is in any case what we have currently in theory, i..e we don't deal with prediction for stacked profiles that will most likely live in FireCrown/new CLcosmo prediction module?). 

There are 2 routes to do so, both shown in the `explore_miscentering_theory.ipynb`
1. rely on the backends to compute the integrals brute-force - the **very** slow option I originally implemented.
2. implement the optimized version from @hsinfan1996, which uses analytical pre-computations to speed up the process.

I like option 2 better ;) 

What do you think?

combet:
Linking @AnnaNiemiec's comment to the PR as well https://github.com/LSSTDESC/CLMM/issues/588#issuecomment-2233487969

Anna, I don't see the update parent_class file with your dblquad implentation pushed to the remote

hsinfan1996:
@AnnaNiemiec Thank you for making dblquad and tplquad work and measuring the performance. I ported the code to `parent_class.py` and did some basic checks on the calculated values. Feel free to test them more on the performance side.

m-aguena (approved):
It all looks good. Thanks for all the hard work! (Just remember to clean the merge comment because it is huge)

marina-ricci:
Hello! I'm taking a look at this PR. There are 3 notebooks related to miscentering, could you explained which one is doing what? Also, having a bit more explanation in the notebook would help to udnerstand what is happening.

combet:
Thanks @marina-ricci - actually, the `explore_miscentering_theory.ipynb` is just where we tried different things to identify what to actually implement in the code. I'm reluctant to delete it completely but it's not meant to be looked at by users (it's not called `demo_...`). Should we move it somewhere?

The only notebooks relevant to this PR are the 2 'historical' theory demo notebooks, where we added a line and plot to show the new functionality in modeling.

Then, there's a mass fitting with miscentering notebook in the mass fitting folder that shows that miscentering will yield an underestimation of the mass if not accounted for. But again, this is not related to that PR.

PR #624, opened 2024-07-15
feat: Add coordinate_system parameter to compute_tangential_and_cross_components and _compute_lensing_angles functions

caioolivv:
Add option to choose coordinate system of ellipticity components.

### DIFF
--- a/clmm/dataops/__init__.py
+++ b/clmm/dataops/__init__.py
@@ -99,7 +99,8 @@ def compute_tangential_and_cross_components(
     shear2: array
         The measured shear (or reduced shear or ellipticity) of the source galaxies
     coordinate_system: str, optional
-        Coordinate system of the ellipticity components. Must be either 'celestial' or 'euclidean'.
+        Coordinate system of the ellipticity components. Must be either 'celestial' or
+        euclidean'. See https://doi.org/10.48550/arXiv.1407.7676 section 5.1 for more details.
         Default is 'euclidean'.
     geometry: str, optional
         Sky geometry to compute angular separation.
@@ -367,7 +368,8 @@ def _compute_lensing_angles_flatsky(
     dec_source_list: array
         Declinations of each source galaxy in degrees
     coordinate_system: str, optional
-        Coordinate system of the ellipticity components. Must be either 'celestial' or 'euclidean'.
+        Coordinate system of the ellipticity components. Must be either 'celestial' or
+        euclidean'. See https://doi.org/10.48550/arXiv.1407.7676 section 5.1 for more details.
         Default is 'euclidean'.
 
     Returns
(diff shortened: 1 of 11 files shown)


caioolivv:
@marina-ricci hey, I think the PR is ready for a review!

clmm/dataops/__init__.py
```diff
@@ -376,12 +389,16 @@ def _compute_lensing_angles_flatsky(ra_lens, dec_lens, ra_source_list, dec_sourc
         phi[angsep == 0.0] = 0.0
     else:
         phi = 0.0 if angsep == 0.0 else phi
+    if coordinate_system == "sky":
+        phi = np.pi - phi
```
hsinfan1996, line 393: `phi = np.arctan2(deltay, deltax)` should do the work. (note: for coordinate_system == "pixel", the calculation is `phi = np.arctan2(deltay, -deltax)`)

caioolivv:
Since @hsinfan1996 mentioned that it was better to keep things consistent between _compute_lensing_angles_flatsky and _compute_lensing_angles_astropy, it may be better to revert the last commit.

hsinfan1996:
> Since @hsinfan1996 mentioned that it was better to keep things consistent between _compute_lensing_angles_flatsky and _compute_lensing_angles_astropy, it may be better to revert the last commit.

Yeah, I am sorry. I was thinking that the calculation would be easier by flipping the sign of `deltax` for the falt-sky version. However, we only use the astropy version right now, so it may be better to keep the calculations consistent. i will let other reviewers to decide which practice is better.

marina-ricci:
I asked @caioolivv to change the notebooks and add an example in the mock_data notebook for the two definitions. @caioolivv let me know when the PR is ready for review !

caioolivv:
Everything's ready @marina-ricci !

clmm/galaxycluster.py
```diff
@@ -35,6 +37,9 @@ class GalaxyCluster:
         Redshift of galaxy cluster center
     galcat : GCData
         Table of background galaxy data containing at least galaxy_id, ra, dec, e1, e2, z
+    coordinate_system : str
```
marina-ricci, line 40: You have to specify which one is the default value.

clmm/support/mock_data.py
```diff
@@ -156,6 +157,9 @@ def generate_galaxy_catalog(
         The number density of galaxies (in galaxies per square arcminute, from z=0 to z=infty).
         The number of galaxies to be drawn will then depend on the redshift distribution and
         user-defined redshift range.  If specified, the ngals argument will be ignored.
+    coordinate_system : str, optional
```
marina-ricci, line 160: same here, specify the default  value

marina-ricci: Also, could you add a reference for these definitions? Like a paper?

marina-ricci (changes requested):
Thank you @caioolivv for this important PR! Since the coordinate system is an important but often missed subtlety, we should clearly highlight it. Hence I suggest : 
- to add a references ( like papers) for the definitions in the code (in comments) and in the notebooks ;
- to highlights the differences between the cordinate systyems and profiles in `demo_dataops_functionality.ipynb`
- to expand also the `demo_mock_cluster.ipynb` notebook.

caioolivv:
Ready for review @marina-ricci

marina-ricci:
@caioolivv I think there is  a typo in `demo_dataops_functionality` notebook that is breaking the code. "coordinate system"=piru

caioolivv:
Sorry @marina-ricci , was testing the paramenter validation and forgot to go back to the original value.

marina-ricci:
Ok, everything seems good now !

marina-ricci (approved):
It lokks good thank you Caio!!!!

Issue #635, opened 2024-07-22
Problem with coordinate implementation


marina-ricci:
We rencentlty merged a PR to specify the coordinate system for the shear, but it lacks some functionalities. 
First, it should be exaplined in details in a notebook, so that this bug would have been found. E.g : showing a profile taken with an incorect cordinate system and showing all intermediate steps. 

At the moment, the keyword is missing from compute_tangential_and_cross_components`. You can pass it to the  `Galaxy Cluster` object, but it does not imply that it will be used (this should be avoided). 

Also, we might consider adding a simple way to convert between the two coordinate, once compute_tangential_and_cross_components` is run (not having to rerun all steps).

(Note : I discovered this while trying to implement in TXPipe and seeing no changes).

caioolivv:
Hey @marina-ricci, just to be sure I understand what's the problem: `GalaxyCluster.compute_tangential_and_cross_components()` should have an option to pass in the coordinate system or should it explain that `GalaxyCluster.coordinate_system` will be used (or a combination of both)?

Also I agree there should be a function to convert between them. Maybe `GalaxyCluster.convert_coordinate_system()`? In that case, should you call it and then call `GalaxyCluster.compute_tangential_and_cross_components()` again or just calling `GalaxyCluster.convert_coordinate_system()` should already compute new tangential and cross components?

Finally, there's the `examples/demo_coordinate_system_datasets.ipynb` notebook from another PR which shows the problems of using another coordinate system. Is that what you meant? In that case we could reference it in another notebook?

marina-ricci:
Hi @caioolivv ! I did not see that the examples/demo_coordinate_system_datasets.ipynb notebook was merged to main. I'm not sure it should stay here in it's current form. 

These are my suggestions : 

- [ ] add warning message when GalaxyCluster (or compute_tangential_xxx, see second point) object is initiated without specifying the coordinate system. Like "no coordinate system provided, taking euclidean as default"
- [ ] put coordinate system as a keyword for `compute_tangential_and_cross_components`, from data_ops or as a GalaxyCluster method. Maybe not necessary to have it as a GalaxyCluster keyword because it may be confusing as it is not necessarily propagated. 
- [ ] create a simple notebook to highlight this. 

I created a branch with a notebook to show the problem (it's not fully documented but should be understandable)

PR #636, opened 2024-07-30
Issue/635/Modify coordinate system implementation

marina-ricci:
This PR aims at resolving issues explained in issue  #635.

Copied from the issue : 

These are my suggestions : 

- [x] add warning message when GalaxyCluster (or compute_tangential_xxx, see second point) object is initiated without specifying the coordinate system. Like "no coordinate system provided, taking euclidean as default"
- [x] put coordinate system as a keyword for `compute_tangential_and_cross_components`, from data_ops or as a GalaxyCluster method. Maybe not necessary to have it as a GalaxyCluster keyword because it may be confusing as it is not necessarily propagated. 
- [x] create a simple notebook to highlight this.


m-aguena:
@caioolivv would you like to take a look at this issue?

caioolivv:
Hi @m-aguena, yes! The last month has been a bit chaotic, but I'll start working on this this week.

caioolivv:
There's another unnoticed bug: calling `compute_tangential_and_cross_components()` repeatedly flips the sign of `e2` when it should stay the same.

caioolivv:
Hi @marina-ricci and @ts485 ! In the demo notebook, there still seems to be a problem in DC2. Is that the same problem you've noticed?

tomomis86:
Hi @caioolivv ,

Was your problem the path to the data directory or something else? If the latter, could you tell me what the problem is? My problem is simply that Cluster-related stages in TXPipe does not run with DC2 anymore.

caioolivv:
> Hi @caioolivv ,
> 
> Was your problem the path to the data directory or something else? If the latter, could you tell me what the problem is? My problem is simply that Cluster-related stages in TXPipe does not run with DC2 anymore.

The problem is in the profile: neither coordinate system creates a profile with a lensing signal.

PR #640, opened 2024-09-20
add Paper_v2.0/paper_formating.py

### LINKED ISSUE #639: Make folder for paper2.0 figures
Create a folder for notebooks/scripts that make paper 2.0 figures

m-aguena:
closes issue #639


marina-ricci:
Could we use this PR to also render all plots? It's annoying to not see them once there are in main.

marina-ricci:
Yes. Run the notebooks with rendered plots (once they are merged). Also we should update the documentation because not all example notebooks are there.

Issue #643, opened 2024-09-24
Computation of PDF integral


m-aguena:
In the current computation of the weights for pdfs with shared bins:
```
        mask = (pzbins >= zmin) * (pzbins <= zmax)
        z_grid = pzbins[mask]
        pz_matrix = np.array(pzpdf)[:, mask]
        kernel_matrix = kernel(z_grid)

    return simpson(pz_matrix * kernel_matrix, x=z_grid, axis=1)
```

the precision is limited to the pdf bin sizes (`pzbins`). This can have a significant impact on the integral. For instance:

![image](https://github.com/user-attachments/assets/d1af10c4-a4ed-4d51-9ac9-b6c72d24ea0c)
 
and `dz=0.05` is similar to what we might get from the PZ WG. So I propose to interpolate the pdf in finer bins before integration

PR #644, opened 2024-09-26
Issue/401/use qp+fix zinteg

### LINKED ISSUE #401: Add p(z) to GalaxyCluster object
This issue is for integrating [qp](https://github.com/LSSTDESC/qp) with CLMM, which would enable #126 and #396. The most basic version of this would involve a new `GalaxyCluster.pz` attribute that is a `qp.Ensemble` object (with one element/row/galaxy, in this case) with its parameters stored in `GalaxyCluster.galcat`. Though qp already has its own data format, it's essentially an Astropy table, just like the `clmm.GCData` object, so it should be possible to do this efficiently, particularly with help from @eacharles.

m-aguena:
Using `qp` for handling PDFs. The main changes are:

- The interpolation of the PDFs on a finer grid for integration in `_integ_pzfuncs` is done by `qp`.
- Adding the possibility of having `quantiles` as the PDF information:
  - Added this option in `generate_galaxy_catalog`
  - Makes the `GCData` object convert the quantiles into a PDF grid with `GCData.get_pzpdfs`, which is used in the integration mentioned above.

Closes #401 and #643


hsinfan1996 (commented):
1. Should we add `qp` to `requirements.txt` and the Requirements section in `README`?

2. It seems to me that the imports in some of the files are not compliant with PEP8, but since `pylint` doesn't complain about C0411 / wrong-import-order, it is not a big deal. I am just wondering.

examples/test_pz_integral.ipynb
combet, line 1: This new notebook is very useful to see the various behaviors, but the markdown text needs to be corrected/expanded to help the user understand better what is happening. At the moment, the markdown seems to copied/pasted from another demo notebook and gives little information as to what is actually happening.

combet: Would be good to add a legend in the figure comparing the pdz computed for the 3 types (individual, shared and quantiles)

combet: Also, maybe add some explanation text for the last figure of the notebook. I find trick to understand what is actually shown.

combet: And why is the qp integral performing worse than the clmm-implemented ones but for the "quantiles" case where qp appears to perform better?

combet (changes requested):
Thank you @m-aguena. It's great to see qp and the quantile option available! 

I've tested the code and everything appears to be working well. I just have comments about the new notebook showing the integral with qp versus clmm, etc. At the moment, I don't think they are enough or clear enough markdown explanations to easily understand what this is doing.

examples/test_pz_integral.ipynb
m-aguena: I added more explanation to the notebook

m-aguena: I think that `qp` is performing worse than `clmm` because it kind of used a type like integral with the precision set by the separation of the zbins of the input provided, and in `clmm` we are refining the grid for the integral.

In the case of using quantiles, I think `qp` might be outperforming `clmm` because it is actually using the statistical meaning of quantiles while `clmm` recreates an interpolated PDF and integrates it, thus loosing information.

clmm/gcdata.py
```diff
@@ -83,6 +88,12 @@ def _str_pzpdf_info(self):
                 np.set_printoptions(edgeitems=5, threshold=10)
                 out += " " + str(np.round(self.pzpdf_info.get("zbins"), 2))
                 np.set_printoptions(**default_cfg)
+            elif out == "quantiles":
+                np.set_printoptions(formatter={'float': "{0:g}".format}, edgeitems=3, threshold=6)
+                out += " " + str(self.pzpdf_info["quantiles"])
+                out += " - upacked with zgrid : " + str(
```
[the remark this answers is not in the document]
m-aguena: Thanks @hsinfan1996 ! I could not figure out how to add `qp` properly to `requirements.txt`. You're welcome to give it a try

hsinfan1996:
I added qp (`qp-prob`) to both `requirements.txt` and `environment.yml`. It would be great if some conda users can check if `environment.yml` works.

INSTALL.md
```diff
@@ -56,9 +56,10 @@ Note, you may choose to install some or all of the ccl, numcosmo, and/or cluster
 Now, you can install CLMM and its dependencies as
 
 ```bash
-    pip install numpy scipy astropy matplotlib
+    pip install numpy scipy astropy matplotlib healpy
     pip install pytest sphinx sphinx_rtd_theme
     pip install jupyter  # need to have jupyter notebook tied to this environment, you can then see the environment in jupyter.nersc.gov
+    git clone https://github.com/LSSTDESC/qp.git; cd qp; python setup.py install; cd - # Install qp package
```
hsinfan1996, line 62: Issuing `python setup.py install` is deprecated. Maybe we can change this part to `pip install git+https://github.com/LSSTDESC/qp.git` for those who don't want to clone the repo.

hsinfan1996 (approved):

combet (approved):
Thanks for adding more explanations to the notebook. All looks good to me know!

PR #646, opened 2024-10-03
Fix some minor problems in boost calculation and improve docstring

combet:
- prevent rscale from being an optional keyword with default value (1000) as it *must* have the same units as rvals. I think it's safer for the user to have to provide a value.
- make corresponding changes 
     - [x] in utils.py
     - [x] in boost notebooks
     - [x] in test_utils 
- Add equations to the docstrings to have a nice rendering in the documentation 
- Rename functions `correct_sigma_with_boost*` as the boost will generally be applied to DeltaSigma or shear and not Sigma.

### DIFF
--- a/clmm/utils/boost.py
+++ b/clmm/utils/boost.py
@@ -49,9 +49,11 @@ def compute_nfw_boost(r_proj, r_scale, boost0=0.1):
             / (2 * np.lib.scimath.sqrt(radicand))
         )
 
-        return np.nan_to_num(finternal, copy=False, nan=1.0).real
+        return np.nan_to_num((1 - finternal) / radicand, copy=False, nan=1.0 / 3.0).real
+        # return np.nan_to_num(finternal, copy=False, nan=1.0).real
 
-    return 1.0 + boost0 * (1 - _calc_finternal(r_norm)) / (r_norm**2 - 1)
+    return 1.0 + boost0 * _calc_finternal(r_norm)
+    # return 1.0 + boost0 * (1 - _calc_finternal(r_norm)) / (r_norm**2 - 1)
 
 
 def compute_powerlaw_boost(r_proj, r_scale, boost0=0.1, alpha=-1):
--- a/clmm/utils/boost.py
+++ b/clmm/utils/boost.py
@@ -3,11 +3,11 @@
 import numpy as np
 
 
-def compute_nfw_boost(rvals, rscale, boost0=0.1):
-    r"""Computes the boost factor at radii rvals using a parametric form 
+def compute_nfw_boost(r_proj, r_scale, boost0=0.1):
+    r"""Computes the boost factor at radii r_proj using a parametric form
     following that of the analytical NFW excess surface density.
 
-    Setting :math:`x = R/R_s` 
+    Setting :math:`x = r_{\rm proj}/r_{\rm scale}`
 
     .. math::
         B(x) = 1 + B_0\frac{1-F(x)}{(x)^2 -1}
@@ -23,20 +23,20 @@ def compute_nfw_boost(rvals, rscale, boost0=0.1):
 
     Parameters
     ----------
-    rvals : array_like
+    r_proj : array_like
         Radii
-    rscale : float
-        Scale radius in same units as rvals
+    r_scale : float
+        Scale radius in same units as r_proj
     boost0: float, optional
         Boost factor normalisation
 
     Returns
     -------
     array
-        Boost factor at each value of rvals
+        Boost factor at each value of r_proj
     """
 
-    r_norm = np.array(rvals) / rscale
+    r_norm = np.array(r_proj) / r_scale
 
     def _calc_finternal(r_norm):
         radicand = r_norm**2 - 1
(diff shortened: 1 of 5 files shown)


clmm/utils/boost.py
```diff
@@ -1,24 +1,39 @@
...
 
 
-def compute_nfw_boost(rvals, rscale=1000, boost0=0.1):
-    """Given a list of `rvals`, and optional `rscale` and `boost0`, return the corresponding
-    boost factor at each rval
+def compute_nfw_boost(rvals, rscale, boost0=0.1):
+    r"""Computes the boost factor at radii rvals using a parametric form 
+    following that of the analytical NFW excess surface density.
+
+    Setting :math:`x = R/R_s`
```
hsinfan1996, line 10: I think it will be clearer to write `x = rvals/rscale`, so poeple know where the arguments are in the equation. We can also rename the argument to `r_proj`, so it is consistent with the theory module.

clmm/utils/boost.py
```diff
@@ -39,25 +54,27 @@ def _calc_finternal(r_norm):
...
 
 
-def compute_powerlaw_boost(rvals, rscale=1000, boost0=0.1, alpha=-1.0):
-    """Given a list of `rvals`, and optional `rscale` and `boost0`, and `alpha`,
-    return the corresponding boost factor at each `rval`
+def compute_powerlaw_boost(rvals, rscale, boost0=0.1, alpha=-1):
+    r"""Computes the boost factor at radii rvals using a power-law parametric form
+
+    .. math::
+        B(R) = 1 + B_0 \left(\frac{R}{R_s}\right)^\alpha
```
hsinfan1996, line 61: Same as the above, `\frac{rvals}{rscale}`

hsinfan1996 (commented):
In the notebook `demo_boost_factors.ipynb`, Is it expected to get `nan` from `print(u.compute_nfw_boost(1000, 1000, boost0=0.1))`?

combet:
> Thank you. In the notebook `demo_boost_factors.ipynb`, is it expected to get `nan` from `print(u.compute_nfw_boost(1000, 1000, boost0=0.1))` (cell 10)?

Ah right, I forgot to remove that line in the notebook... I was checking things out. Indeed, I think that's a problem but the denominator will go to zero when x=1, so that's expected. Question is whether we want to add a warning or something?

hsinfan1996:
For completeness, yes, we should probably add something to deal with the case `R_proj = R_s`. My guess is that the value F(x) at x = R_proj / R_s = 1 is just a constant, so we can fill the value in the output array.

clmm/utils/boost.py
```diff
@@ -1,24 +1,39 @@
...
+
+    .. math::
+        B(x) = 1 + B_0\frac{1-F(x)}{(x)^2 -1}
+
+    where
+
+    .. math::
+        F(x) = \begin{cases} \frac{\arctan \sqrt{x^2 -1 }}{\sqrt{x^2 -1 }} & \text{if $x>1$}, \\
+        \text{1} & \text{if x = 1}, \\
+        \frac{\text{arctanh} \sqrt{1-x^2 }}{\sqrt{1- x^2 }} & \text{if $x<1$}.
```
hsinfan1996, line 20: In the actual calculation (`_calc_finternal` below), the arctan and arctanh is inside a natural log.

combet: I meant to ask who coded up the boost, but my guess is that the complex notation with natural log used in the code ends up being equivalent to explicitely writing things in terms of arctan and arctanh. I half-convinced myself this was the case but didn't spend enough time on this. Would be good to have confirmation

hsinfan1996: Wolfram alpha says they are the same expression. Pretty cool.
https://www.wolframalpha.com/input?i=arctanh%28sqrt%28x%5E2-1%29%29%2Fsqrt%28x%5E2-1%29

clmm/utils/boost.py
hsinfan1996: Just saw that there was an extra space at the end of this line.

clmm/utils/boost.py
```diff
@@ -1,24 +1,39 @@
 """General utility functions that are used in multiple modules"""
+
 import numpy as np
 
 
-def compute_nfw_boost(rvals, rscale=1000, boost0=0.1):
-    """Given a list of `rvals`, and optional `rscale` and `boost0`, return the corresponding
-    boost factor at each rval
+def compute_nfw_boost(rvals, rscale, boost0=0.1):
+    r"""Computes the boost factor at radii rvals using a parametric form
```
hsinfan1996, line 7: An extra space at the end of this line.

hsinfan1996:
We can do the following to deal with the case `R_proj = R_s`:

```
    r_norm = np.array(rvals) / rscale

    def _calc_finternal(r_norm):
        radicand = r_norm**2 - 1

        finternal = (
            -1j
            * np.log(
                (1 + np.lib.scimath.sqrt(radicand) * 1j) / (1 - np.lib.scimath.sqrt(radicand) * 1j)
            )
            / (2 * np.lib.scimath.sqrt(radicand))
        )

        return np.nan_to_num((1 - finternal)/radicand, copy=False, nan=1.0/3.0).real

    return 1.0 + boost0 * _calc_finternal(r_norm)
```

combet:
> We can do the following to deal with the case `R_proj = R_s`:
> 
> ```
>     r_norm = np.array(rvals) / rscale
> 
>     def _calc_finternal(r_norm):
>         radicand = r_norm**2 - 1
> 
>         finternal = (
>             -1j
>             * np.log(
>                 (1 + np.lib.scimath.sqrt(radicand) * 1j) / (1 - np.lib.scimath.sqrt(radicand) * 1j)
>             )
>             / (2 * np.lib.scimath.sqrt(radicand))
>         )
> 
>         return np.nan_to_num((1 - finternal)/radicand, copy=False, nan=1.0/3.0).real
> 
>     return 1.0 + boost0 * _calc_finternal(r_norm)
> ```

Thanks, I implemented that and that took care of the issue!

clmm/utils/boost.py
combet: fixed

combet:
> For completeness, yes, we should probably add something to deal with the case `R_proj = R_s`. My guess is that the value F(x) at x = R_proj / R_s = 1 is just a constant, so we can fill the value in the output array and don't need to show warnings.

Done. And docstrings updated accordingly.

hsinfan1996:
@combet Thank you.

hsinfan1996 (approved):

m-aguena (approved):
Looks good! Thanks @hsinfan1996 for doing the hard part of the review!

PR #649, opened 2024-11-08
Issue/635/Modify coordinate system implementation

caioolivv:
Recreating PR #636 with correct branch name.


caioolivv:
I've noticed I still have to update other notebooks that had coordinate system examples. Working on it right now.

marina-ricci:
@caioolivv is this PR ready for review? If so, could you convert from 'draft' to 'ready to review' ?

caioolivv:
> @caioolivv is this PR ready for review? If so, could you convert from 'draft' to 'ready to review' ?

Hi, its done now! Sorry for the delay

marina-ricci:
Thanks! I was just testing the same thing. 

In the notebook, I suggest adding shapenoise=0.05"` in the source kwargs. Also for the histogram of values, could you change to (for visibility) : 

`f, ax = plt.subplots(1, 2, figsize=(10, 4))

ax[0].hist(cl_euclidean.galcat["et"], bins=50, color="tab:blue", alpha=0.5, label='euclidean')
ax[0].hist(cl_celestial.galcat["et"], bins=50, color="tab:red", histtype='step', label='celestial')
ax[0].hist(cl_wrong.galcat["et"], bins=50, color="tab:orange", alpha=0.5,  histtype='stepfilled', label='incorrect coordinate system')
ax[0].set_xlabel("$\\epsilon_t$", fontsize="xx-large")

ax[1].hist(cl_euclidean.galcat["ex"], bins=50, color="tab:blue", alpha=0.5, label='euclidean')
ax[1].hist(cl_celestial.galcat["ex"], bins=50, color="tab:red", histtype='step', label='celestial')
ax[1].hist(cl_wrong.galcat["ex"], bins=50, color="tab:orange", alpha=0.5,  histtype='stepfilled', label='incorrect coordinate system')
ax[1].set_xlabel("$\\epsilon_x$", fontsize="xx-large")
ax[1].set_yscale("log")

plt.legend()
plt.show()`

marina-ricci:
What worried me is that this bug seemed to have pass the tests...

caioolivv:
Wait, isn't that right though? Mirroring the x-axis consists of doing phi -> pi - phi, which is supposed to change the sign of e_x.

clmm/galaxycluster.py
```diff
@@ -297,7 +289,7 @@ def compute_tangential_and_cross_components(
             dec_lens=self.dec,
             geometry=geometry,
             validate_input=self.validate_input,
-            coordinate_system=self.coordinate_system,
+            coordinate_system=self.galcat.meta["coordinate_system"],
```
marina-ricci, line 292: Need to put this in the documentation of the function.

caioolivv: Resolved

combet:
Hi @caioolivv, @m-aguena, @marina-ricci   - I'm starting the review this PR. Here are a first few things I noticed:
- there's currently a minor conflict to be solved with `main`. 
- the `demo_coordinate_system` notebook is missing two imports for the "HSC Y3" section to work properly. Namely
```
from astropy.io import fits
from scipy import spatial
```

clmm/gcdata.py
```diff
@@ -221,6 +239,30 @@ def update_cosmo(self, cosmo, overwrite=False):
...
+        *args : tuple
+            Components to be converted, must be in data.
+
+        Returns
+        -------
+        None
+        """
+        in_coordinate_system = self.meta["coordinate_system"]
+        self.update_info_ext_valid("coordinate_system", self, coordinate_system, overwrite=True)
+        if coordinate_system != in_coordinate_system:
```
combet, line 259: This is where having an else statement with a warning saying the requested coordinate system is already the one of the catalog and therefore no change is being made to the catalog.

clmm/gcdata.py
```diff
@@ -221,6 +239,30 @@ def update_cosmo(self, cosmo, overwrite=False):
...
+            for col in args:
+                if col in self.colnames:
+                    self[col] = -self[col]
```
combet, line 262: Does this means that if the user mistakenly passes a column that should not be affected by the coordinate system, e.g. redshift column, all redshifts will be transformed with a minus sign? Maybe we need another warning to the user to say, "careful, this operation should only be used on the ellipticity columns" (or something like this?).

clmm/support/mock_data.py
```diff
@@ -486,7 +490,7 @@ def _generate_galaxy_catalog(
     return galaxy_catalog[cols]
 
 
-def _draw_source_redshifts(zsrc, zsrc_min, zsrc_max, ngals):
+def _draw_source_redshifts(zsrc, zsrc_min, zsrc_max, ngals, coordinate_system):
```
combet, line 493: It's a bit counter intuitive that a functions that only draw redshifts needs to know about the coordinate system. I understand because this function is called first to generate the mock data and the initial galaxy catalog comes from it.

An option would be to declare an empty galaxy catalog GCData with the coord system in the metadata, and then, have `_draw_source_redhsifts` adds to that GCData. Similarly to what `_draw_galaxy_positions` or `_compute_photoz_pdfs` do?

In any case, if we keep it like this, `coordinate_system` needs an argument docstring. It's currently missing.

combet (dismissed):
Thank you for this PR! This new functionality appears robust and well documented. The dedicated notebook is very useful. I left a few comments above for your attention. Do not hesitate to ask if something is unclear.

clmm/gcdata.py
caioolivv: Yes, that would happen... I couldn't find a way to prevent this given that columns can be renamed arbitrarily. I will add a warning though.

clmm/gcdata.py
```diff
@@ -65,6 +71,16 @@ def __init__(self, *args, **kwargs):
         APtable.__init__(self, *args, **kwargs)
         metakwargs = kwargs["meta"] if "meta" in kwargs else {}
         metakwargs = {} if metakwargs is None else metakwargs
+        if (
+            "coordinate_system" not in metakwargs
```
m-aguena, line 75: what is the application of `copy_indices` and `copy` here?

clmm/support/mock_data.py
```diff
@@ -44,7 +44,7 @@ def generate_galaxy_catalog(
     pz_bins=101,
     pz_quantiles_conf=(5, 31),
     pzpdf_type="shared_bins",
-    coordinate_system="euclidean",
+    coordinate_system=None,
```
m-aguena, line 47: why was this set to `None`?

clmm/gcdata.py
```diff
@@ -239,45 +239,43 @@ def update_cosmo(self, cosmo, overwrite=False):
...
+            return
+
+        if ellipticity_columns:
+            missing_columns = list(
+                filter(lambda col: col not in self.colnames, ellipticity_columns)
+            )
+            if len(missing_columns) > 0:
+                raise ValueError(f"Components {missing_columns} not found in data. No changes made")
+            for col in ellipticity_columns:
+                self[col] *= -1
```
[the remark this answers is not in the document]
m-aguena: Good point. I will fix it

clmm/support/mock_data.py
```diff
@@ -542,7 +567,10 @@ def _draw_source_redshifts(zsrc, zsrc_min, zsrc_max, ngals):
     else:
         raise ValueError(f"zsrc must be a float, chang13 or desc_srd. You set: {zsrc}")
 
-    return GCData([zsrc_list, zsrc_list], names=("ztrue", "z"))
```
m-aguena, line 545: Why not just adding `coordinate_system` to `_draw_source_redshifts` instead of creating a new  function (`_create_gcdata`) ?

m-aguena (commented):
Great job @caioolivv ! I just left a couple of comments. I still have to check the notebooks and will let you know when is done

m-aguena:
@caioolivv can you save a copy of the data generated by the `demo_coordinate_system_datasets.ipynb` notebook at cc-in2p3 in
```
/sps/lsst/groups/clusters/CLMM/HSC_data_example/
```

and comment it on the notebook?

Thanks

clmm/gcdata.py
m-aguena: @caioolivv I think adding these changes on `__getitem__` is the better approach. As you mentioned, `cosmo` was already being lost when a slice was made. What was the problem with pylint? Depending on the error we can ignore it

m-aguena (approved):
@caioolivv looks good now, you can merge it. Just remeber to update the version and clean the squash and merge message.
Thanks!

Issue #618, opened 2024-02-22
Allow empty radial bins in cluster ensemble


marina-ricci:
When working with cluster ensemble we can add individual profiles to produce a stack. However, if those profiles do not have the same radial bin the stacking step will fail. This is a problem when integrating CLMM in the Cluster pipeline as we provide the same N radial bins for every cluster, but when a radial bin is empty an array N-1 will be returned. For stacking cluster_ensemble requires all profiles with the same radial bins.

combet:
`make_radial_profile` has a keyword `include_empty_bins` that is False by default. Turning it to True when calling that function should resolve the problem (I think).

hsinfan1996:
> `make_radial_profile` has a keyword `include_empty_bins` that is False by default. Turning it to True when calling that function should solve the problem (I think).

It is set to `True` and user is not allowed to change this argument. (line 161 in the `clusteremsemble.py`)

combet:
Ah yes, it is if using the method `make_individual_radial_profile` from the CLMM ClusterEnsemble class. 

I believe @marina-ricci came across the issue when calling CLMM from TXPipe (see [this](https://github.com/LSSTDESC/TXPipe/blob/issue_309_CL_data_vector/txpipe/extensions/clmm/make_ensemble_profile.py)), where the individual profiles are computed using `make_radial_profile` and sequentially added to a ClusterEnsemble object using the `add_individual_radial_profile` method.

hsinfan1996:
> Ah yes, it is if using the method `make_individual_radial_profile` from the CLMM ClusterEnsemble class (but this assumes all clusters background catalog to have been stored in the ClusterEnsemble, which is not feasible if there are too many clusters). 
> 
> 
> 
> I believe @marina-ricci came across the issue when calling CLMM from TXPipe (see [here](https://github.com/LSSTDESC/TXPipe/pull/327#issuecomment-1949342705) and [this](https://github.com/LSSTDESC/TXPipe/blob/issue_309_CL_data_vector/txpipe/extensions/clmm/make_ensemble_profile.py)), where the individual profiles are computed using  GC `make_radial_profile` method and sequentially added to a ClusterEnsemble object using the `add_individual_radial_profile` method. In that case, having `include_empty_bins=False` in the `make_radial_profile` call should fix the issue.
> 
> 
> 
> 

Ah I see. Yes, `include_empty_bins=True` should fix it. Sorry for the misunderstanding.

marina-ricci:
[...]

PR #652, opened 2024-11-18
add warning for empty bins in ClusterEnsemble

m-aguena:
Closes issue #618 , second interaction


marina-ricci:
Just to copy here what we discussed : we want to return NaN instead of 0 when there are empty bins.

m-aguena:
> Just to copy here what we discussed : we want to return NaN instead of 0 when there are empty bins.

done!

m-aguena:
@combet the notebook issue has been fixed

combet:
Thanks @m-aguena - the figure looks good now

Screenshot 2025-03-26 at 09 43 49

May I ask adding a sentence after the figure to comment on the sample covariance having a few empty rows/columns because of empty bins, while this is not a problem for BS and JK methods.

Also, in the output of cell 15, below, is not really explained and we do not see dashed lines in the plot, contrary to what the legend says. I’m actually not sure I understand what this figure aims at showing, so a bit more text would be useful

Screenshot 2025-03-26 at 09 50 12

m-aguena:
> Thank you @m-aguena - I think a bit more comments in the notebooks are needed to help the user understand everything that is shown (see my comments above). Apart from that, everything looks good.

done!

combet (approved):
Thanks @m-aguena - it looks all good to me now!

Issue #657, opened 2025-01-23
API break for NumCosmo>v0.22.0


m-aguena:
After the NumCosmo update (v>0.22.0), the numcosmo backend theory module brakes

PR #659, opened 2025-01-23
fix to numcosmo<=0.22.0

m-aguena:
closes issue #657


hsinfan1996:
NumCosmo v0.23 is compatible, but I don't think it matters in this case.

hsinfan1996 (approved):

PR #661, opened 2025-02-05
Issue/658/api break for numcosmo

caioolivv:


m-aguena (changes requested):
@caioolivv thanks for the work! It seems the changes are working now, but the `Example2_Fit_Halo_Mass_to_Shear_Catalog_NC.ipynb` notebook is still failing:
```
AttributeError: 'NoneType' object has no attribute 'mid'
```

m-aguena:
> > @caioolivv thanks for the work! It seems the changes are working now, but the `Example2_Fit_Halo_Mass_to_Shear_Catalog_NC.ipynb` notebook is still failing:
> > ```
> > AttributeError: 'NoneType' object has no attribute 'mid'
> > ```
> 
> Just fixed it, but that reminds me of two things:
> 
> 1. I've just updated for `numcosmo<=0.24`. Versions 0.25 and up have a different problem that I'll have to think about... in another PR.
> 2. This `Example2_Fit_Halo_Mass_to_Shear_Catalog_NC.ipynb` notebook contains an old unbinned likelihood model that has been completely reworked to a new version we've discussed adding to a fitting module in CLMM.  I think we leave this as is at the moment and edit it only when we decide how to add the new unbinned likelihood method.

Agreed, can you create an issue to remember of revisiting  `Example2_Fit_Halo_Mass_to_Shear_Catalog_NC.ipynb` in the future? Thanks

PR #671, opened 2025-05-22
Fixed deprecated parameter in Numpy function

### LINKED ISSUE #669: Function problem with Numpy 2.x
In the function `numpy.nan_to_num`, the option `copy=False` was deprecated after `Numpy == 2.0`. Just removing this option on the function will not change the functionality.

To reproduce the error:

```
from clmm.utils.boost import compute_nfw_boost


val = compute_nfw_boost(2.0, 1.0)
```
Output:

ValueError: Unable to avoid copy while creating an array as requested.
If using `np.array(obj, copy=False)` replace it with `np.asarray(obj)` to allow a copy when needed (no behavior change in NumPy 1.x).
For more details, see https://numpy.org/devdocs/numpy_2_0_migration_guide.html#adapting-to-changes-in-the-copy-keyword.

eduardojsbarroso:
Pull request to fix issue [669](https://github.com/LSSTDESC/CLMM/issues/669).

### DIFF
--- a/clmm/__init__.py
+++ b/clmm/__init__.py
@@ -27,4 +27,4 @@ from .theory import (
 )
 from .utils import compute_radial_averages, convert_units, make_bins
 
-__version__ = "1.16.2"
+__version__ = "1.16.3"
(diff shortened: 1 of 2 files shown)


m-aguena (approved):
looks good to me

PR #672, opened 2025-05-23
Issue/670/fix miscentering

hsinfan1996:
Fix miscentering calculations not taking `r_proj` of float type.

### DIFF
--- a/clmm/theory/miscentering.py
+++ b/clmm/theory/miscentering.py
@@ -191,7 +191,7 @@ def integrate_azimuthially_miscentered_surface_density(
     aux_args : list
         Auxiliary arguments used in the integral
     extra_integral : bool
-        Additional dimention for the integral
+        Additional dimension for the integral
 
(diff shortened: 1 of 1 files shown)


hsinfan1996:
@eduardojsbarroso Can you check if the issue is fixed?

eduardojsbarroso:
> @eduardojsbarroso Can you check if the issue is fixed?

Hi Hsin,
yes I just installed CLMM from this branch and run the same code again. Everything seems to be working now, thank you. In my point of view, we can merge this to the main branch.

hsinfan1996:
Hi @eduardojsbarroso ,
Thank you for the prompt reply. I will have @m-aguena do a quick review on this PR.

m-aguena (changes requested):
I think this makes the output always an array, right?
Can you make it a float if the input is a float? (This should be made on the public functions of `CLMModeling`)

PR #702, opened 2025-12-10
Issue/701/fail_missing_backend_tests [version:1.16.11]

### LINKED ISSUE #701: Having CLMM tests skip if a backend is not available means some bugs might be left unseen
For example, until recently we had a hard check on CCL's version in order to prevent us from trying to use an unsupported version of CCL. Still, in the CI we installed the latest version of CCL, which of course meant that every test that required CCL as a backend was skipped because of the hard check. That means that for the last few months we weren't testing anything with CCL and had no way to know this.

caioolivv:
## Description

Fail tests if backends are not available. Tests should be ran at all times.

## Main changes
- Chnage missing backend erros from pytest.skip to pytest.fail

## Checklist

Besides passing all CI checks and coverage is at 100%, make sure you also checked the following items
(check details in [CONTRIBUTING](https://github.com/LSSTDESC/CLMM/blob/main/CONTRIBUTING.md)).

### For developers

- [x] **Notebooks:** notebooks related to this PR have been updated and all notebooks can run correctly.
- [x] **Build the documentation:** All documentation builds correctly.

### For reviewers

- [ ] **Notebooks:** notebooks related to this PR have been updated and all notebooks can run correctly.
- [ ] **Build the documentation:** All documentation builds correctly.

### For developers (part 2)

After the PR has been approved by two reviewers:

- [x] Update the code version in `clmm/__ini__.py`.
- [ ] Keep only relevant points in the squash and merge commit message.
- [x] If any dependencies have been altered, update `environment.yml`, `pyproject.toml`, `INSTALL.md`, and `README.md`. A maintainer should also be notified to change the requirements on conda-forge.
- [x] Update `clmm/theory/_ccl_supported_versions.py` and `clmm/theory/ccl.py` if pyccl's version constraints have been altered.


m-aguena (changes requested):
@caioolivv thanks for repporting this. I guess the issue here is that we don't want to force the user to have all backends installed, but we do want the CI to run with all backends. Do you think there is a way we can achieve this?

